Issue #33, opened 2026-03-15
[OPH Task] papers.compact.e.27-upgrade-gauge-electroweak-closure-from-consistency-check-to-prediction: 27. Upgrade gauge/electroweak closure from consistency check to prediction


muellerberndt:
This issue is managed by `automation/oph_completion/oracle_task_pipeline.py`.
Update the task tree or runtime state instead of editing the issue body directly.

## Status
- Pipeline status: `pending`
- Kanban target status: `Backlog`
- Attempts: `0`
- Completed runs: `0`

## Task
- Task ID: `papers.compact.e.27-upgrade-gauge-electroweak-closure-from-consistency-check-to-prediction`
- Parent: `papers.compact.e`
- Primary file: `reverse-engineering-reality/paper/recovering_relativity_and_standard_model_structure_from_observer_overlap_consistency_compact.tex`
- Dependencies:
  - `papers.compact.e.26-derive-rg-matching-and-threshold-structure-from-oph`

## Summary
Make the gauge couplings and electroweak outputs genuine forward outputs from the OPH inputs rather than consistency checks on a calibrated branch.

## Prompt Focus
Rewrite the quantitative gauge and electroweak branch so it reads as prediction rather than post-calibrated validation.

## Acceptance Criteria
- Gauge closure and electroweak outputs are promoted from calibrated checks to forward predictions or are explicitly demoted where that cannot yet be justified.
- Dependencies, imported premises, and remaining blockers are stated explicitly rather than hidden in packaging or rhetoric.
- All configured companion surfaces stay aligned with the updated status of the primary surface.

## Target Files
- `reverse-engineering-reality/paper/recovering_relativity_and_standard_model_structure_from_observer_overlap_consistency_compact.tex`
- `reverse-engineering-reality/paper/tex_fragments/PAPER.tex`
- `reverse-engineering-reality/paper/deriving_the_particle_zoo_from_observer_consistency.tex`

PR #143, opened 2026-04-03
fix: demote D10 electroweak closure claims (issue #33)

rikkaishere5594:
- Sync README, README_FR, the synthesis paper, the particle paper, and the technical supplement to describe D10 as a Phase II calibration branch rather than theorem-grade electroweak closure.
- Keep the public W/Z rows as forward-emitted D10 branch outputs and treat the exact frozen electroweak pair as compare-only validation.
- Leave the compact note and PAPER fragment unchanged because upstream main already carries the demoted D10 framing recorded in the issue 33 artifact.


muellerberndt:
Closed after manual reconciliation into the current frontier. The useful parts were ported selectively: D10 remains a Phase II calibration branch, the frozen exact W/Z pair stays compare-only, and the active public electroweak surface is preserved as a closed target-free theorem on that branch. This PR is therefore superseded by the synced paper-side reconciliation rather than merged verbatim.

Issue #68, opened 2026-03-15
[OPH Task] papers.reality.b.08-prove-observable-level-confluence-in-the-quantum-setting: 8. Prove observable-level confluence in the quantum setting


muellerberndt:
This issue is managed by `automation/oph_completion/oracle_task_pipeline.py`.
Update the task tree or runtime state instead of editing the issue body directly.

## Status
- Pipeline status: `pending`
- Kanban target status: `Backlog`
- Attempts: `0`
- Completed runs: `0`

## Task
- Task ID: `papers.reality.b.08-prove-observable-level-confluence-in-the-quantum-setting`
- Parent: `papers.reality.b`
- Primary file: `reverse-engineering-reality/paper/reality_as_consensus_protocol.tex`
- Dependencies:
  - `papers.reality.b.07-lift-the-record-layer-to-operator-algebras`

## Summary
Show that even when microscopic representatives differ by gauge or sector labels, all physical observables converge to the same overlap-consistent values.

## Prompt Focus
Extend confluence from microscopic representatives to physical observables in the quantum lift of the consensus theorem stack.

## Acceptance Criteria
- Observable-level confluence is proved in the quantum setting even when microscopic representatives differ by gauge or sector labels.
- Dependencies, imported premises, and remaining blockers are stated explicitly rather than hidden in packaging or rhetoric.
- All configured companion surfaces stay aligned with the updated status of the primary surface.

## Target Files
- `reverse-engineering-reality/paper/reality_as_consensus_protocol.tex`
- `reverse-engineering-reality/paper/recovering_relativity_and_standard_model_structure_from_observer_overlap_consistency_compact.tex`
- `reverse-engineering-reality/paper/tex_fragments/TECHNICAL_SUPPLEMENT.tex`

PR #190, opened 2026-04-10
feat: paper: prove observable-level confluence on the quantum lift

cryptoverse-cyber:
#68

This PR closes the theorem gap in
`papers.reality.b.08-prove-observable-level-confluence-in-the-quantum-setting`.

The consensus stack already proved fixed-cutoff confluence on the declared quotient-local repair
carrier and already showed that gauge-invariant observables descend with that carrier, but it did
not yet state the sharper quantum claim the task needed: even when microscopic representative
lifts differ by gauge or sector relabelings, all physical observables on the fixed-cutoff
physical algebra converge to the same overlap-consistent values.

This change fills that bridge directly:
- the consensus paper now defines the fixed-cutoff physical observable algebra on the quantum lift
  and proves observable-level confluence on the physical quotient, not just representative-level
  normal-form uniqueness;
- the live consensus supplement now adds the missing bridge that representatives of the same
  quotient-local glued state induce the same state on the fixed-point / quotient-local observable
  algebra, including the central sector projectors and record observables built from them;
- the compact, master, and observer-facing summary surfaces now say the same thing at the same
  claim tier: D1 fixes the quotient normal form and the terminal expectation functionals on the
  declared physical observable algebras, while a unique microscopic representative remains out of
  scope.

Net effect: the paper stack no longer leaves observable objectivity implicit or over-tied to
microscopic representative uniqueness. It now states the stronger defensible result: on the
declared fixed-cutoff quantum lift, physical observables are schedule-independent and
representative-independent even when microscopic lifts differ by gauge relabelings globally or by
sector/higher-gauge relabelings inside one declared quotient-local glued state.


Armstrong24:
Went through all six files carefully. The math checks out and this correctly closes open problem #68.

The new Theorem thm:obsconfluence with Definition def:physobs and Corollary cor:obsglue chains naturally from thm:confluence + thm:gauge. Both branches are handled well — the T_Σ-orbit argument on the noncentral branch is the right approach. The updated D1 table entries, the new observable-level confluence bullet in reality_as_consensus_protocol.tex, and the "Resolution:" rename are all clean and consistent throughout.

One small thing, the proof of cor:obsglue on the noncentral branch would read better with an explicit cross-reference to Definition def:physobs so it's clear which algebra the physical observables are pulled from. Not a blocker at all, just a readability note.

Overall looks good to merge.

Issue #47, opened 2026-03-15
[OPH Task] papers.observers.c.07-close-the-cosmological-constant-screen-capacity-package: 7. Close the cosmological-constant / screen-capacity package


muellerberndt:
This issue is managed by `automation/oph_completion/oracle_task_pipeline.py`.
Update the task tree or runtime state instead of editing the issue body directly.

## Status
- Pipeline status: `pending`
- Kanban target status: `Backlog`
- Attempts: `0`
- Completed runs: `0`

## Task
- Task ID: `papers.observers.c.07-close-the-cosmological-constant-screen-capacity-package`
- Parent: `papers.observers.c`
- Primary file: `reverse-engineering-reality/paper/tex_fragments/TECHNICAL_SUPPLEMENT.tex`
- Dependencies:
  - `papers.observers.c.06-close-the-black-hole-information-evaporation-package`

## Summary
Unify the local null-blindness argument with the global capacity law and make the monograph's cosmological-capacity statements theorematic.

## Prompt Focus
Present the cosmological-constant and screen-capacity branch as one theorem stack instead of a split local-plus-global story.

## Acceptance Criteria
- The cosmological-constant and screen-capacity package is presented as one theorematic framework with explicit local-versus-global scope.
- Dependencies, imported premises, and remaining blockers are stated explicitly rather than hidden in packaging or rhetoric.
- All configured companion surfaces stay aligned with the updated status of the primary surface.

## Target Files
- `reverse-engineering-reality/paper/tex_fragments/TECHNICAL_SUPPLEMENT.tex`
- `reverse-engineering-reality/paper/recovering_relativity_and_standard_model_structure_from_observer_overlap_consistency_compact.tex`
- `reverse-engineering-reality/paper/tex_fragments/PAPER.tex`

PR #194, opened 2026-04-10
fix: sync README with the cosmological-capacity boundary (issue #47)

rikkaishere5594:
- Rewrite the README de Sitter/gravity delivery bullet so it states the same local/global split already present in the paper stack
- Say explicitly that local null data fix the Einstein branch only up to a metric term and that `Lambda` is fixed only after the global screen-capacity identification
- Keep the public README aligned with the current issue-47 claim boundary without changing the underlying theorem surface


rikkaishere5594:
Closing this PR to rework issue #47 against the intended task artifact.

PR #196, opened 2026-04-11
paper/compact: tighten the BW cap-modular theorem boundary

cryptoverse-cyber:
This PR closes the remaining paper-surface gap behind
`papers.compact.c.08-replace-the-bw-cap-modular-proof-sketch-with-a-full-theorem`. #14 

The live compact paper already carries the right theorem-tier structure: geometric modular flow on
caps is stated as a conditional scaling-limit theorem on the extracted prime geometric subnet, the
\(2\pi\) normalization is fixed explicitly, and the remaining UV burden is sharpened to the
realized transported cap-local system plus the transported fixed-local-collar Markov/faithfulness
datum beneath the derived carried-collar schedule. The remaining issues were companion-surface and
notation mismatches that still weakened or blurred that theorem boundary.

This change tightens those surfaces in four places:
- the compact paper introduction no longer describes the current BW lane as merely a proof-sketch
  surface; it now states that longer arguments are written in compressed proof form;
- the master synthesis fragment now keeps the BW theorem chain on the same explicit carrier
  \(\omega_\infty^{\mathrm{geo},C}\) all the way through the theorem proof and the cap-first-law
  handoff, instead of drifting to the looser \(\omega_\infty^C\) shorthand at the key formulas;
- the top-level observers synthesis paper now states the Lorentz/null-modular/Einstein lane as a
  conditional scaling-limit branch with the same explicit \textbf{T2} plus extraction-and-rigidity
  boundary used by the compact paper, instead of summarizing it as a vague continuum branch with
  unnamed technical premises;
- the observer synthesis summary now names the actual external burdens on the Lorentz/null-modular
  side: the \textbf{T2} scaling-limit scope clause, the theorem-local realized geometric cap-pair
  extraction plus ordered cut-pair rigidity on the extracted prime geometric subnet, and the
  fixed-cap stationarity step.

Net effect: the compact paper, the master synthesis fragment, and the observer synthesis surface
now describe the BW/cap-modular theorem at the same claim tier, with the same carrier, and with
the same explicit UV scaffold and remaining blockers.

Issue #14, opened 2026-03-15
[OPH Task] papers.compact.c.08-replace-the-bw-cap-modular-proof-sketch-with-a-full-theorem: 8. Replace the BW / cap-modular proof sketch with a full theorem


muellerberndt:
This issue is managed by `automation/oph_completion/oracle_task_pipeline.py`.
Update the task tree or runtime state instead of editing the issue body directly.

## Status
- Pipeline status: `completed`
- Kanban target status: `Under audit`
- Attempts: `2`
- Completed runs: `0`

## Task
- Task ID: `papers.compact.c.08-replace-the-bw-cap-modular-proof-sketch-with-a-full-theorem`
- Parent: `papers.compact.c`
- Primary file: `reverse-engineering-reality/paper/recovering_relativity_and_standard_model_structure_from_observer_overlap_consistency_compact.tex`
- Dependencies:
  - `papers.compact.b.07-derive-quasi-local-propagation-and-refinement-stability-internally`

## Summary
Prove geometric modular flow on caps directly from OPH, including the 2pi normalization and a controlled refinement limit, without an independent cap-isotropy or Euclidean-regularity premise.

## Prompt Focus
Upgrade the cap-modular geometry branch from proof sketch to theorem and make its normalization and refinement limits explicit.

## Acceptance Criteria
- Geometric modular flow on caps is proved directly from OPH with explicit normalization and refinement control.
- Dependencies, imported premises, and remaining blockers are stated explicitly rather than hidden in packaging or rhetoric.
- All configured companion surfaces stay aligned with the updated status of the primary surface.

## Target Files
- `reverse-engineering-reality/paper/recovering_relativity_and_standard_model_structure_from_observer_overlap_consistency_compact.tex`
- `reverse-engineering-reality/paper/tex_fragments/PAPER.tex`
- `reverse-engineering-reality/paper/tex_fragments/TECHNICAL_SUPPLEMENT.tex`

PR #200, opened 2026-04-13
fix: paper/compact: align the D3-D5 Lorentz/gravity boundary across companion surfaces

cryptoverse-cyber:
This follow-up closes the remaining companion-surface gap behind
`papers.compact.c.08-replace-the-bw-cap-modular-proof-sketch-with-a-full-theorem`. From https://github.com/FloatingPragma/observer-patch-holography/pull/196#issuecomment-4230612837. Closing #14 

The live compact paper already carries the right theorem-tier structure: geometric modular flow on
caps is stated as a conditional scaling-limit theorem on the extracted prime geometric subnet, the
\(2\pi\) normalization is fixed explicitly, and the remaining UV burden is sharpened to the
realized transported cap-local system plus the transported fixed-local-collar Markov/faithfulness
datum beneath the derived carried-collar schedule. The remaining issues were companion-surface and
ledger-alignment mismatches that still weakened or blurred that theorem boundary.

This change tightens those surfaces in four places:
- the compact paper introduction no longer describes the current BW lane as merely a proof-sketch
  surface; it now states that longer arguments are written in compressed proof form;
- the master synthesis fragment now keeps the BW theorem chain on the same explicit carrier
  \(\omega_\infty^{\mathrm{geo},C}\) all the way through the theorem proof and the cap-first-law
  handoff, instead of drifting to the looser \(\omega_\infty^C\) shorthand at the key formulas;
- the top-level observers synthesis paper now states the Lorentz/null-modular/Einstein lane with
  the same D3/D4/D5 burden split used by the compact ledger: D3 carries \textbf{T2} plus the
  theorem-local geometric cap-pair extraction and ordered cut-pair rigidity; D4 keeps the
  inherited-strip and exact-or-controlled Markov hypotheses, the same-half-line local
  modular-Hamiltonian input, and the separate bounded-interval projective branch explicit; D5
  keeps fixed-cap stationarity, the locally Lorentzian \(d=4\) scaling regime, and the
  tensor-upgrade side conditions explicit;
- the observer synthesis summary now matches that same D3/D4/D5 split instead of compressing the
  whole Lorentz/null-modular/Einstein lane into a D3-only BW boundary.

Net effect: the compact paper, the master synthesis fragment, the observer root, and the observer
synthesis surface now describe the Lorentz/null-modular/Einstein chain at the same claim tier,
with the same BW carrier, and with the same explicit D3/D4/D5 external burdens and remaining
blockers.


muellerberndt:
Thanks, I'm merging this, but it also triggers a whole lot of changes throughout other papers, book, textbooks etc so I'll do an immediate follow-up commit that syncs everything.

PR #203, opened 2026-04-14
Close the physical Majorana phase surface

### LINKED ISSUE #154: [OPH Task] particles.majorana.01-close-majorana-phase-surface: Close the physical Majorana-phase surface
## Goal
Emit the physical Majorana-phase data on the active OPH neutrino surface instead of leaving the Majorana structure implicit inside auxiliary continuation machinery.

## Why this issue exists
The current corpus closes the weighted-cycle neutrino theorem lane for masses, splittings, and PMNS-type observables, but it does not yet expose final physical Majorana phases as public constants.

## Success rule
- The public neutrino surface states whether physical Majorana phases are emitted, unphysical, or still absent.
- If emitted, the dependency chain from the current neutrino branch to those phases is explicit and target-free.
- If not emitted, the exact remaining blocker is stated sharply rather than absorbed into generic neutrino language.

## Dependencies
Likely downstream of the current neutrino theorem pair and any remaining physical attachment / phase-readout closure objects.

MarioPoneder:
## Summary
- add an explicit shared-basis transport of the repaired weighted-cycle Majorana matrix through `U_e_left`
- emit theorem-grade physical `alpha21^(Maj)` and `alpha31^(Maj)` via canonical Takagi congruence on that transported branch
- surface the new Majorana phase rows in the particle status outputs and constants ledger with focused regression coverage

## Path To The Solution
This issue did not close on the first obvious route.

A direct readout from PMNS columns is not an honest physical Majorana readout, because arbitrary intermediate column rephasings leave the oscillation observables unchanged while moving `alpha21^(Maj)` and `alpha31^(Maj)`.

That means the phase pair has to be fixed only after canonical Takagi congruence of the emitted complex symmetric weighted-cycle matrix, with the electron-row gauge `U_e1 in R_{>0}`.

Even with that stable Takagi readout, public promotion still should not rest on an implicit row-order assumption. The final solution therefore rebuilds the physical PMNS path explicitly on the closed shared same-label basis by transporting the repaired weighted-cycle branch itself:

- `M_shared = U_e_left^* M_wc U_e_left^dagger`
- `U_nu_shared = U_e_left U_wc`
- `U_PMNS = U_e_left^dagger U_nu_shared`

On that anchored shared-basis transport branch, the repaired weighted-cycle PMNS surface is recovered exactly, and canonical Takagi congruence then emits one theorem-grade physical Majorana phase pair.

This closes the public neutrino Majorana surface on the shared-basis weighted-cycle transport branch and avoids the earlier overclaim that a PMNS-only or implicit-basis route was already sufficient.

Closes #154.

## Test Plan
- [x] `python -m pytest code/particles/neutrino/test_neutrino_weighted_cycle_shared_basis_representation.py code/particles/neutrino/test_physical_majorana_phase_theorem.py code/particles/test_results_status_neutrino_comparison_rows.py code/particles/test_results_status_majorana_surface.py`
- [x] regenerate `code/particles/runs/neutrino/neutrino_weighted_cycle_shared_basis_representation.json`
- [x] regenerate `code/particles/runs/neutrino/neutrino_physical_majorana_phase_theorem.json`
- [x] regenerate `code/particles/RESULTS_STATUS.md`, `code/particles/results_status.json`, and `code/particles/runs/status/status_table_forward_current.json`
- [x] independently verify the shared-basis transport identities and emitted Majorana values

## Notes
- I did not run the full disposable runtime workspace build end-to-end because `code/particles/compute_current_output_table.py` expects an external ancillary particle tree that is not present in this local workspace.


muellerberndt:
LGTM

Issue #56, opened 2026-03-15
[OPH Task] papers.observers.e.16-close-the-problem-of-time-claim-via-modular-time: 16. Close the problem-of-time claim via modular time


muellerberndt:
This issue is managed by `automation/oph_completion/oracle_task_pipeline.py`.
Update the task tree or runtime state instead of editing the issue body directly.

## Status
- Pipeline status: `pending`
- Kanban target status: `Backlog`
- Attempts: `0`
- Completed runs: `0`

## Task
- Task ID: `papers.observers.e.16-close-the-problem-of-time-claim-via-modular-time`
- Parent: `papers.observers.e`
- Primary file: `reverse-engineering-reality/paper/observers_are_all_you_need.tex`
- Dependencies:
  - `papers.observers.d.15-decide-the-status-of-the-critical-superstring-lift`

## Summary
Turn the monograph's problem-of-time entry into an actual theorem or corollary from modular flow and observer algebras rather than a conceptual gloss.

## Prompt Focus
Use the wrapper closure table and gravity surfaces to make the problem-of-time claim either theorematic or explicitly downgraded.

## Acceptance Criteria
- The monograph's problem-of-time claim is either a theorem or corollary from modular flow and observer algebras or is explicitly downgraded.
- Dependencies, imported premises, and remaining blockers are stated explicitly rather than hidden in packaging or rhetoric.
- All configured companion surfaces stay aligned with the updated status of the primary surface.

## Target Files
- `reverse-engineering-reality/paper/observers_are_all_you_need.tex`
- `reverse-engineering-reality/paper/recovering_relativity_and_standard_model_structure_from_observer_overlap_consistency_compact.tex`
- `reverse-engineering-reality/paper/tex_fragments/TECHNICAL_SUPPLEMENT.tex`

muellerberndt:
Closed by the manual integration of PR #253 paper-side content on main. The problem-of-time surface is stated as a conditional BW/geometric-branch observer-relative time reading, not as an operational-clock theorem or global solution. The compact paper, PAPER fragment, Observers appendices, and compact supplement carry the downgrade/alignment in release r1405. Book edits from the PR were intentionally not merged.

PR #253, opened 2026-05-01
fix: qualify modular time claim (issue #56)

rikkaishere5594:
- Defines the problem-of-time surface as a conditional BW/geometric-branch observer-relative time reading, not an operational-clock theorem.
- Synchronizes the compact paper, PAPER fragment, supplement port, and appendix status table to the downgraded claim tier.
- Qualifies book-wide modular-time and clock language with candidate / observer-relative / BW-branch caveats.

Tested: patch --dry-run -p2; git diff --check; targeted overclaim phrase sweep.


muellerberndt:
Integrated manually into main without the book edits, as requested. The paper-side modular-time qualifiers from this PR were adapted to the current refactored sources in commit 023faed, and the final rebuilt paper bundle is release r1405.

Issue #26, opened 2026-03-15
[OPH Task] papers.compact.d.20-internalize-fermions-and-chirality: 20. Internalize fermions and chirality


muellerberndt:
This issue is managed by `automation/oph_completion/oracle_task_pipeline.py`.
Update the task tree or runtime state instead of editing the issue body directly.

## Status
- Pipeline status: `pending`
- Kanban target status: `Backlog`
- Attempts: `0`
- Completed runs: `0`

## Task
- Task ID: `papers.compact.d.20-internalize-fermions-and-chirality`
- Parent: `papers.compact.d`
- Primary file: `reverse-engineering-reality/paper/recovering_relativity_and_standard_model_structure_from_observer_overlap_consistency_compact.tex`
- Dependencies:
  - `papers.compact.d.19-replace-the-compact-gauge-reconstruction-sketch-with-a-full-proof`

## Summary
Complete the fermionic or super-Tannakian branch and derive chiral light matter internally instead of leaving chirality attached to a selector problem or open-question marker.

## Prompt Focus
Replace bosonic-only structural strength with an internal fermionic and chiral matter derivation or cleanly demote the unsupported claims.

## Acceptance Criteria
- Fermions and chirality are derived internally or the unsupported chiral-closure language is explicitly removed from the structural core.
- Dependencies, imported premises, and remaining blockers are stated explicitly rather than hidden in packaging or rhetoric.
- All configured companion surfaces stay aligned with the updated status of the primary surface.

## Target Files
- `reverse-engineering-reality/paper/recovering_relativity_and_standard_model_structure_from_observer_overlap_consistency_compact.tex`
- `reverse-engineering-reality/paper/tex_fragments/PAPER.tex`
- `reverse-engineering-reality/paper/tex_fragments/GAUGE_GROUP_DERIVATION.tex`

muellerberndt:
Closed by the manual integration of PR #254 on main. The compact paper and PAPER fragment now limit D7 to bosonic compact-gauge reconstruction and keep fermions/chirality on the explicit realized matter-package premise rather than promoting them to a completed super-Tannakian reconstruction theorem. The wording is resolved against the current color-first N_c=3, N_g=3 theorem stack in release r1405.

PR #254, opened 2026-05-02
fix: demote unsupported chirality closure (issue #26)

rikkaishere5594:
- Limits the recovered gauge theorem boundary to bosonic compact-gauge reconstruction.
- Presents the Standard Model quotient, exact hypercharge lattice, and N_g/N_c counting as realized-branch results under explicit matter-package and admissibility premises.
- Keeps the unresolved fermionic/super-Tannakian reconstruction gap visible instead of claiming completed internal chirality reconstruction.

Verification: git diff --check; source grep for explicit bosonic/realized-branch boundary language.
Not tested: TeX build.


muellerberndt:
Integrated manually into main after resolving against the current color-first Nc/Ng theorem stack. The chirality-closure demotion is included in commit e9d8a3e, and the final rebuilt paper bundle is release r1405.

Issue #279, opened 2026-05-12
Paper cleanup: observers status table marks closed BW and compact-gauge lifts as open


muellerberndt:
## Problem
The observers appendices and synthesis fragments contain status language that marks the continuum BW lift and the refinement-limit compact-gauge lift as open. That contradicts the current theorem chain: the support-visible BW scaling theorem supplies the continuum modular/geometric lift, and the compact-gauge branch is supplied on the zero-obstruction bosonic realized branch by the transportability, fixed-cutoff sector-category, refinement/fiber descent, sector-colimit, and realized witness theorems.

## Confirmed paper-source hits
- `paper/tex_fragments/OBSERVERS_APPENDICES.tex:212` says fixed-cutoff theorem packages exist, but the continuum BW lift on the extracted geometric subnet and the refinement-limit transportable compact-gauge lift remain open.
- `paper/tex_fragments/OBSERVERS_SYNTHESIS_SECTIONS.tex:249` to `:251` lists compact heat-kernel lift, global consensus analysis, refinement-stable universality, and continuum/gravity lift as external burdens on the reference-architecture lane.

## Required fix
- Update the UV completion / microscopic uniqueness row so it says the remaining boundary is microscopic representative uniqueness, not the BW lift or the compact-gauge lift.
- In the synthesis section, separate local reference-architecture scope from corpus-level theorem status. Do not list continuum/gravity lift as an open burden of the OPH corpus.
- If compact heat-kernel lift or refinement-stable universality remains a real D10 or architecture-lane dependency, state it in that narrow quantitative branch only, not as a recovered-core or BW/gauge gap.
- Regenerate the observers paper and generated markdown after the source edit.

## Acceptance criteria
- The observers paper does not say the continuum BW lift remains open.
- The observers paper does not say the refinement-limit transportable compact-gauge lift remains open on the realized zero-obstruction bosonic branch.
- Remaining future-program rows are limited to genuine unfinished lanes such as microscopic uniqueness, strong CP, charged-lepton source landing, critical-string continuations, or other explicitly unfinished particle/continuation surfaces.

PR #290, opened 2026-05-12
fix: Align observers status surfaces with the closed BW and compact-gauge theorem branches

cryptoverse-cyber:
This PR tightens the observers paper status language so it matches the current OPH theorem stack.

Problem/gap:

- the observers appendix status table still said the continuum BW lift and the refinement-limit transportable compact-gauge lift remained open;
- that wording conflicted with the current companion corpus, where the support-visible BW scaling theorem closes the modular/geometric Lorentz branch and the realized zero-obstruction bosonic compact-gauge branch is carried by the transportability, fixed-cutoff sector-category, refinement/fiber descent, sector-colimit, and realized-witness theorems;
- the synthesis fragments needed narrower wording cues so the remaining engineering burdens stay local to the fixed-cutoff microphysics lane rather than sounding like recovered-core gaps.

What this PR does:

- rewrites the `UV completion / microscopic uniqueness` row in the observers appendix so the future-program boundary is unique microscopic representative selection, not a missing BW or compact-gauge lift;
- tightens the observers synthesis fragment so the compact heat-kernel lift is explicitly local to the microphysics lane while the BW / geometric-modular / local-Einstein closure stays identified as a companion recovered-core surface;
- tightens adjacent observers synthesis bullets so the noncentral-gluing and gap summaries do not imply a missing geometric scaling branch or unresolved compact-gauge lift elsewhere in the same paper;
- regenerates the observers paper markdown export after the source edit.

Net effect:

- the observers paper no longer reports the closed BW branch as open;
- the observers paper no longer reports the realized zero-obstruction bosonic compact-gauge branch as an unresolved lift;
- remaining future-program language stays focused on genuine unfinished lanes rather than stale recovered-core status.

### DIFF
--- a/paper/tex_fragments/OBSERVERS_APPENDICES.tex
+++ b/paper/tex_fragments/OBSERVERS_APPENDICES.tex
@@ -209,7 +209,7 @@ companion-paper theorem surfaces.
 Claim & Status & Boundary \\
 \midrule
 Quantum gravity consistency & Proved corollary of earlier theorems & Lorentz and Jacobson-type Einstein branches follow from the support-visible BW scaling theorem together with the stated null-bridge and fixed-cap conditions. \\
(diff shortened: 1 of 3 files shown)


cryptoverse-cyber:
#279

Issue #304, opened 2026-06-05
Proof packet: Boundary-Fiber Identifiability


muellerberndt:
Created from the missing-proof agenda in `docs/missing_proofs.md`.

## Completed generic theorem

For a finite quotient patch net with a preserved boundary or sector map, repair completeness, local diamond on the quotient, and singleton consistent fibers modulo gauge, equal protected observations determine the same observer-facing normal form.

- [x] State observation-preservation, completeness, singleton-fiber, and weak-normalization hypotheses.
- [x] Prove boundary identification modulo gauge if and only if cross-source endpoint uniqueness.
- [x] Separate same-source confluence, cross-source reconstruction, weak normalization, and all-schedule liveness.
- [x] Supply finite positive, coarse-boundary, and observation-leak examples.
- [x] Formalize the substrate-neutral theorem in Lean.
- [x] Propagate the generic claim boundary through the paper and learning surfaces.

The theorem and proof package live in *Observation-Determined Normal Forms: Stability, Obstructions, and Refinement in Constraint and Rewrite Systems* and `Lean/ObserverPatchHolography/Proofs/ObservableNormalForms/`.

## Scope of this closed issue

This issue closes the generic boundary-fiber theorem. The theorem does not define the physical OPH boundary map or prove its injectivity modulo gauge on the Einstein-tower domain.

The concrete application and its refinement-uniform inverse modulus are tracked in #577 under the Einstein branch-entry umbrella #503.

muellerberndt:
Tracking note (2026-06-24): mutable progress, dependencies, blockers, and completion state for this task belong in this GitHub issue and the DAG2 project state. Paper and book prose should state stable claim boundaries, theorem premises, receipt requirements, or claim classes only; they should not carry project-status ledgers or work-progress language.

muellerberndt:
Consistency pass (2026-06-24): local papers, books, cosmology branch, particle artifacts, and simulator docs/code should carry stable claim boundaries, receipt labels, and publication scope only. Mutable progress, blockers, dependencies, and completion state for this item stay in GitHub/DAG2, not in paper prose or local status trackers.

qizwiz:
the observer-facing uniqueness half of this goes through.. machine-checked in Lean, CI-green.

[...]

PR #385, opened 2026-07-02
feat(LEAN): Rule 90 carrier — Hfib discharged on an information-set boundary + repair no-go (#304)

qizwiz:
the Hfib hypothesis from #304 ("for the declared B, C_b is a singleton mod gauge") discharged on the first carrier where it's non-degenerate.. the linear case, where it's exactly erasure correction.

one Rule 90 step as a two-patch carrier: edge-consistency ⟺ the bottom row is the Rule 90 image of the seed ⟺ a valid CA diagram. rule 90 is linear over F2, so the consistent set is literally a code and "B identifies the fiber" becomes "B contains an information set".

machine-checked, standard axioms only (propext, Classical.choice, Quot.sound.. no sorry, no native_decide):

- `rule90_Hfib_good`: reading bottom cells {0,1} identifies the fiber. exactly the `Hfib` binder of `boundary_fiber_observer_unique`, instantiated.. it plugs in directly
- `rule90_Hfib_bad_fails`: reading {0,2} fails. same carrier, same dynamics, coarser read.. it reads the CA-redundant bit twice and erases the only non-redundant cell
- `rule90_gauge_nontrivial`: seeds (0,0,0) and (1,0,1) are observably identical.. a ker(Rule90) pair, the unobservable part of the seed (the exact gauge = kernel iff is next-packet material)
- `rule90_no_frustrationFree_repair`: NO local repair satisfying H1-H3 exists on this carrier at all. every Rule90 image has equal outer bits, so any target row that breaks that is out of reach: H2 forces a fire, H1 pins it to the seed, H3 demands the impossible

this is exactly the obligation from your last comment on #304: B injective on the consistent quotient C / gauge. `rule90_Hfib_good` is that statement, discharged.. and the bad-boundary theorem shows the property is about WHICH cells B reads, not how many.

the no-repair theorem is why I think your injectivity reframe is not just cleaner but forced, at least for rigid carriers like this one: the repair-selector route is provably closed here, while the injectivity route goes through. I have a sketch suggesting this generalizes ("no holography without frustration": under H1-H3 on locally-simple carriers, every Hfib boundary must carry full bulk information.. no compression) — that's the next packet if it survives.

next packet either way: the general multi-patch carrier, B injective on the consistent set ⟺ interior forced mod gauge.. the constraint-code version, where the consistent set is generated by local parity checks and "boundary determines interior" becomes literal erasure decoding.

builds green on top of b0e186e (8252 jobs, x86-64 linux). axiom audit in the build log.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


muellerberndt:
I got two audit findings here:

1. [Rule90.lean (line 59)](/tmp/oph-pr385/LEAN/ObserverPatchHolography/Rule90.lean:59) overstates the no-repair result. rule90_no_frustrationFree_repair proves there is no single-site H1/H2/H3 frustration-free local repair on this carrier. It does not rule out every repair route, e.g. transactional multi-patch repair, a different repair-site carrier, relaxed H2, or a repair law outside the LocalRepairDynamics binder. The docstrings at lines 59-62 and 190-200 should say “the H1-H3 local-repair route is closed on this carrier,” not “the repair route is provably closed.”

2. The PR adds a new #304-targeting Lean module but does not update the Lean proof index/README, despite the repo’s own update protocol in [PROOF_INDEX.md (line 111)](/tmp/oph-pr385/LEAN/PROOF_INDEX.md:111). This leaves the public Lean status stale: [README.md (line 15)](/tmp/oph-pr385/LEAN/README.md:15) still says the primitives are 10 sorry-bearing placeholders, and [PROOF_INDEX.md (line 9)](/tmp/oph-pr385/LEAN/PROOF_INDEX.md:9) still says 0% / 0 discharged. At minimum, add rows for Rule90.lean and update the proof-debt/count language to reflect the post-#384/#385 state.

qizwiz:
thanks, both fair.

1. softened the two docstrings (Rule90.lean:59-62, 190-200) to "the H1-H3 local-repair route is closed on this carrier," and spelled out what it does NOT close (transactional/multi-patch repair, a different repair-site carrier, relaxed H2). you're right that "provably closed" overstated it.. the theorem only kills the single-site H1-H3 route.

2. added a #304 section to PROOF_INDEX.md with rows for the 5 Rule90 theorems + a completion-summary line, plus a README bullet. left Prop 4.2 / Def 4.1 at 0% since this is a carrier-level witness, not a Prop 4.2 item.. didn't want to inflate the count.

pushed as c8a831e.

Issue #519, opened 2026-07-11
[Publication audit] Reconcile falsifiability, applications, framing, and numeric errata


muellerberndt:
## Public-surface reconciliation
All reader-facing claims must match the claim registry and executable evidence class.

## Required corrections
- Remove the `Cheap Desktop Fusion Energy`, `500 W-class home-scale energy cell`, hoverbike, and hoverboard claims from `docs/APPLICATIONS.md` unless a named receipt supports each statement.
- Reconcile README, book, paper, website, and application language with registry status.
- Use conditional-likelihood terminology for the IBM packet; do not call it a Bayes factor or evidence for OPH over unrestricted QM.
- Make the claim validator green before publication.

## Acceptance
A repository-wide scan finds no unsupported quantitative application claim, stale status label, fixed live-issue count, or reader-facing claim stronger than its evidence path.

muellerberndt:
Closing on the acceptance scan executed with the coherence pass landed in commit 05c12b4a (branch fable_pass). Dispositions: (1) docs/APPLICATIONS.md no longer carries the desktop-fusion, 500 W home-cell, or hoverbike/hoverboard claims as promises; all four use cases are labeled design concepts under an explicit no-receipt evidence banner, and the lift concept states that no effect has been measured and that the force law itself is conditional. (2) README, book, paper, and application language reconcile against the registry: the claim validator passes (48 claims, 15 owner papers), the reader style checker passes, and the survival-integration ledgers pass. (3) The IBM packet language is conditional-likelihood throughout; the only Bayes-factor phrasing in the corpus is the explicit disclaimer that the comparisons are not Bayes factors. (4) A repository scan finds no unsupported quantitative application claim, no hardcoded live-issue count, and no stale status label; manifest membership is derived from the build-tool source set rather than a fixed count. Numeric errata handled in the same pass: the comparison pixel is now labeled P_C at every site that printed comparison-branch values under P_star, both electroweak bridge capacities carry their pixel-branch tags, and the withdrawn raw-basis Z near-hit is recorded in the particle paper.

PR #583, opened 2026-07-18
docs/APPLICATIONS.md: remove unsupported Fusion and Hoverbike claims (#519)

WGlynn:
Implements the first Required Correction of #519: remove the unsupported application claims from `docs/APPLICATIONS.md`.

## What changed
- Removed **Section 1 "Cheap Desktop Fusion Energy"** (including the "cheap 500 W-class home-scale energy cell" target).
- Removed **Section 4 "Hoverbikes And Hoverboards"**.
- Removed the "Fusion cell" and "Hoverbike or hoverboard" rows from the Common Pattern table.
- Renumbered the two retained tracks and updated the intro count ("Four use cases" to "Two use cases") and the closing line ("energy, compute, intelligence, and lift" to "compute and intelligence").

Net: 5 insertions, 89 deletions; no other files touched.

## Why these two
Per the acceptance criterion I scanned for a named receipt backing each statement and found none: there is no fusion or hoverbike/lift entry under `claims/` or in the `code/` receipts. The retained sections are kept deliberately:
- **SHA-256d compute** cites `Photonic Fixed-Point Consensus for SHA-256d Proof of Work` and is already self-hedged ("the measured prototype loses to random search under a matched exact-verification budget").
- **AGI** cites `Thinking as Patch-Net Fixed-Point Search` and related papers.

## Scope note
This PR edits only `docs/APPLICATIONS.md`, as the issue directs. "fusion"/"hover" still appear in some papers and book chapters (for example `extra/hacking-the-simulation-anti-gravity-exploit/`, several `book/` chapters); those are your claims surface and appropriately hedged where I checked, so I left them untouched and am flagging them here rather than editing.

If you would rather **qualify** these two sections with an explicit "no receipt yet" disclaimer instead of removing them (the alternative you noted in #519), say so and I will switch to that.

### DIFF
--- a/docs/APPLICATIONS.md
+++ b/docs/APPLICATIONS.md
@@ -8,43 +8,13 @@ drives repair. Stable outcomes become records. OMEGA is the lab version of
 that idea: physical chambers, labeled ports, control software, and exact
 checks.
 
-Four use cases follow. All four are low-cost implementation tracks under
+Two use cases follow. Both are low-cost implementation tracks under
 active development. The links point to public material in this repository or
 public OPH pages. The OMEGA hardware visualization is at
 [omega.floatingpragma.io](https://omega.floatingpragma.io/). The local compute loop diagram is
 [assets/omega/omega-compute-loop.svg](../assets/omega/omega-compute-loop.svg).
 
-## 1. Cheap Desktop Fusion Energy
-
-The first use case is a closed-loop acoustic fusion cell. Think of a small
-water-based cavity driven as an OPH patch. It tracks pressure modes, charge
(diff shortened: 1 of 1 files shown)


muellerberndt:
Yes,. I'd rather qualify these applications rather than removing them. They'll be decided IRL in experiments.

WGlynn:
Makes sense. Since these get decided in real experiments, qualifying them is the better call than removing them outright. I would rather not put words in your mouth on the physics, so I will leave the qualifying wording to you as the author. Closing this removal PR so you can qualify the Fusion and Hoverbike sections directly. Thanks for the steer.

PR #555, opened 2026-07-17
Make the mandatory scientific suite collectable from a clean clone

WGlynn:
## What this does

From a fresh clone, the mandatory pytest suite (`pytest.ini`: `testpaths = code`) cannot be collected. Running `python -m pytest --collect-only code` in a clean environment at `153725c` stops with 25 collection errors. 24 of them trace to a single undeclared dependency, `mpmath`, which is imported by 39 files under `code/` but appears in no requirements file. `sympy` is undeclared in the same way (5 files).

This PR makes the mandatory lane collectable and documents the one command path, without touching any physics or scientific content.

## Changes

- **`requirements.txt`** (new, root): pins the core dependencies for the mandatory lane, including the previously undeclared `mpmath` and `sympy`. Versions are the set verified from a clean clone on CPython 3.12.
- **`code/ibm_quantum_cloud/conftest.py`** (new): gates the optional IBM / Qiskit lane out of default collection behind `OPH_RUN_IBM=1`. This mirrors the existing `OPH_RUN_LEGACY_D10` opt-in in `code/particles/conftest.py`. Without it, the optional lane fails collection on any clone that does not have the IBM extras installed.
- **`REPRODUCE.md`** (new): documents the clean-clone command path and the opt-in extras.

## Verification

At `153725c`, clean venv, core deps only:

| step | before | after |
| --- | --- | --- |
| `pytest --collect-only code` exit | 2 (interrupted) | 0 |
| collection errors | 25 | 0 |
| tests collected | 794 | 918 |

`python tools/check_claim_registry.py` continues to pass (28 claims, 15 owner papers). The optional lane is still runnable with `OPH_RUN_IBM=1 python -m pytest code/ibm_quantum_cloud`.

This addresses the "make the mandatory top-level verification command pass from a fresh clone" and "optional IBM/Qiskit lane must be an explicit extra" acceptance criteria in #507. A full independent clean-clone reproduction, including the mandatory-lane pass/fail counts, is posted on #553.

## Scope

Collection health only. This does not change any theorem, receipt, or numerical claim, and it does not attempt to fix individual scientific test failures. The pins are deliberately minimal and can be relaxed to ranges if you prefer.

Refs #507, #553.

Issue #542, opened 2026-07-11
[Paper audit] Eliminate theorem-surface TeX warning debt


muellerberndt:
## Scope

The TeX corpus must build under the canonical source-derived commands with warning-free output or a documented narrow warning budget. Review verification includes all eight core papers, all nine root `extra/*.tex` papers, all seven unpublished `cosmology/*.tex` papers, and the reader-facing book.

## Implemented state

The build covers eight core, nine extra and seven cosmology roots, builds the book, retains the exact logs consumed by the warning checker, and applies a source-anchored allowlist. Overfull boxes, missing references, missing glyphs, and font-substitution defects are rejected. Publication CI rebuilds the papers and book twice and compares their hashes, and it rejects tracked PDFs that drift from a clean source rebuild. The committed book is compared with a scratch rebuild. Newly generated untracked or ignored paper PDFs fail instead of being missed by `git diff`; six temporary Git fixtures exercise the actual workflow step. Cosmology-only edits trigger the workflow and all 24 paper PDFs participate in the two-pass hash comparison.

The book's 25 SVG inputs are bound by source and rendering SHA-256 to canonical PDF assets. A normal clean-clone build validates and stages those exact bytes, keeping host librsvg, Pango, Cairo, and font-stack differences outside the reproducibility boundary.

## Acceptance

- [x] The compact paper, main paper, supplements, all root `extra/*.tex` and `cosmology/*.tex` papers, and book build under the canonical local commands.
- [x] Underfull warnings are eliminated or covered by a documented narrow source-anchored budget.
- [x] Overfull boxes, missing references, missing glyphs, and font substitutions fail closed.
- [x] CI records and enforces the warning budget, tracked-artifact drift, and second-pass byte identity.
- [x] Warning-only layout fixes do not lengthen claims or change theorem status.
- [x] The final Phase 0 source commit passes the Ubuntu Publication Build Gate and the Ubuntu/Windows Mandatory Suite from clean checkouts.

## Original closure evidence (historical)

- Source commit: `6166cc304fcd1aca86ea79262e8de2b24e2952ed`
- Ubuntu Publication Build Gate: https://github.com/FloatingPragma/observer-patch-holography/actions/runs/30145576220
- Ubuntu and Windows Mandatory Suite: https://github.com/FloatingPragma/observer-patch-holography/actions/runs/30145476385

## Release boundary

[...]

PR #585, opened 2026-07-18
#542 first pass: TeX typesetting hardening + underfull warning-budget allowlist

WGlynn:
First pass on the theorem-surface TeX warning debt (#542). Built the full corpus with tectonic 0.16.9 and worked the two gate classes the checker enforces.

## What this PR does (safe, source-only)
1. **Uniform hardening.** Adds `microtype` and `\emergencystretch=3em` to the 11 papers that were missing one or both. Preamble-only, +24 lines, **zero prose or claim changes** (the diff is pure insertions). This drops overfull boxes corpus-wide from **43 to 30** and lowers underfull badness.
2. **Underfull warning budget.** Adds `paper/build_warning_allowlist.json`, the file `check_build_warnings.py` already looks for but which did not exist. It documents a per-source-file underfull budget (164 boxes) using the checker's robust matcher (`source_file` + `badness_max` + `max_count`, no line/badness pinning, so it survives tectonic-version drift). With it, the checker reports **0 unexplained underfull**.

Verification (Windows, tectonic 0.16.9): every paper builds; `check_build_warnings.py` shows 0 unexplained underfull; the only remaining failures are overfull boxes.

## What this PR does NOT close (honest boundary)
**30 overfull boxes remain**, so the checker still fails on those. They are wide, unbreakable content sitting inside theorem-dense prose, where the correct fix is an author layout judgment (break the display, resize, or reword a non-claim identifier) that I did not want to make blindly on the physics:

- `deriving_the_particle_zoo...`: 10 (long `.json` identifier lists; inline math value tuples like `(80:..., 91:...)`)
- `screen_microphysics...`: 6 (wide math rows, worst 324pt; a few empty-excerpt wide boxes = tables/displays)
- `fine_structure...`: 5 (wide math rows `; ; ;`, worst 77pt; one table cell)
- `reality_as_consensus_protocol`: 1 (87pt math chain)
- `thinking_as_patch_net...`: 1 (81pt inline file path `(../docs/...)`)
- `recovering_..._compact`: 1 (19pt inline dimension tuple)
- `OBSERVERS_APPENDICES` fragment: 1 (194pt overfull **vbox** = a too-tall page block)
- `observers`, `machine_checked` (1.3pt), `PARTICLE_..._SUPPLEMENT` (2), `FINITE_COVARIANT...` (0.4pt): small residuals

Most of these fall into three fixable shapes once you decide the house convention: (a) long identifier/path lists to `\seqsplit`/`\path` or a `sloppypar`, (b) wide inline math tuples broken or set as a display, (c) wide display rows wrapped in `\resizebox`/`adjustbox`. I can do that pass on your say-so, or you may prefer to place the breaks yourself. Underfull budget can then be tightened as those get eliminated.

Relates to #542. No CI workflow added (per your note on #555).


WGlynn:
Heads up on CI: this PR edits `paper/**/*.tex`, so the Claim Registry workflow's "claim-surface changed without touching the registry surface" gate will flag it. That is a false positive here. The only `.tex` changes are preamble additions (`\usepackage{microtype}` and `\setlength{\emergencystretch}{3em}`); no claim, theorem, or prose text is touched (the diff is pure insertions, 0 deletions). Happy to adjust if you would rather scope the gate to exclude preamble-only edits, but I did not want to touch the registry surface for a typesetting change.

muellerberndt (changes requested):
The preamble hardening is useful and the partial scope is honest, but the proposed allowlist is too broad to merge as a warning gate. Matching only  +  +  with  allows any underfull warning in that source up to the count ceiling; the 111-warning entry is effectively a blanket waiver and can hide entirely new layout regressions. Please either (a) keep the microtype/emergency-stretch changes and remove the allowlist from this PR, or (b) narrow each budget with stable excerpt/category anchors (and tighter badness bounds where practical) and add a regression test showing that an unseen underfull warning in the same source remains unexplained. No claim-registry content change is requested for preamble-only edits; we can handle that workflow false positive separately.

muellerberndt:
Formatting clarification for the review above: the current entries match only log, source_file, kind, badness_max, and max_count. In particular, badness_max 10000 is the full possible range. The requested narrowing is unchanged.

WGlynn:
Thanks, that is a fair objection. I took path (b) and narrowed the budget so it can no longer act as a blanket waiver (dc53cfd8).

What changed:

1. The allowlist no longer carries per-file blanket counts. Each entry now anchors ONE known underfull to its stable `source_file` + line-range + an observed badness band (`badness_min`/`badness_max`), with `max_count` per anchored location. The old 111-count entry is gone. A warning at a different location, or a badness outside the anchored band, is treated as unexplained and fails the gate.

2. Regression tests added (all green locally, 3/3):
   - `test_unseen_underfull_in_allowlisted_source_is_unexplained` - a new underfull in an allowlisted source but at a different line-range is NOT absorbed (this is exactly the regression you asked for).
   - `test_same_location_but_worse_badness_is_unexplained` - a worse badness at the same line-range falls outside the band and stays unexplained.
   - `test_known_underfull_is_allowed` - the anchored known underfull remains budgeted.

3. `gen_warning_allowlist.py` regenerates these anchored entries from a clean build log, so the budget stays honest rather than hand-maintained.

The one red check is `validate` (claim-registry): the 12 preamble-only `.tex` edits trip the "claim-surface changed without a registry edit" rule. That is the workflow false positive you flagged as separate, so I left the claim-registry gate untouched rather than paper over it here. Happy to defer to however you want to handle it.

muellerberndt (approved):
Addressed in the follow-up commit: the broad per-file budget was replaced by source-line/badness anchors, with a regression test proving unseen warnings remain unexplained. Local targeted tests pass.

Issue #514, opened 2026-07-11
[Data integrity] Manifest-derived release validation (credential rotation waived)


muellerberndt:
## Security and data-integrity boundary

PDG reference integrity is covered by four passing focused tests. Release-manifest membership is derived from the registered source set, not a fixed entry count.

**Owner decision 2026-07-23:** credential rotation is waived. `code/ibm_quantum_cloud/.api-key` remains gitignored and untracked; the owner accepts the local-checkout presence of the key as the standing arrangement. The credential acceptance items are closed by that decision, not by rotation.

## Re-audit evidence, 2026-07-25

The manifest generator and validator now:

- derive the complete TeX and registered non-TeX paper set;
- require every registered non-TeX source to have its declared PDF;
- bind each paper ID to its exact source-derived PDF path rather than accepting a valid PDF under another ID;
- reject missing, extra, stray, cross-swapped, malformed, and hash-mismatched artifacts;
- use deterministic manifest metadata; and
- rely on a clean source rebuild plus `git diff --exit-code` to reject a committed PDF/manifest pair whose bytes do not match its source.

The 14 focused manifest tests pass. An independent adversarial re-audit also cross-swapped real paper PDFs and updated their manifest digests; the deterministic rebuild restored the source-derived bytes and the tracked-artifact drift gate rejected the substitution.

## Acceptance

- [x] Reference values carry external provenance, units, and integrity tests.
- [x] CI runs registry and manifest checks.
- [x] Credential handling is waived by owner decision; no rotation is required.
- [x] Manifest validation derives the expected paper set and exact ID-to-path mapping from the registered sources.
- [x] Missing, extra, stray, cross-swapped, malformed, and stale artifacts fail closed.
- [x] A clean rebuild rejects committed publication bytes that do not match their sources.

muellerberndt:
State of the data-integrity and credential items:

Present in the repository:
- `code/particles/data/particle_reference_values.json` carries PDG 2026 API provenance per entry (source label, `summary_id`, `request_timestamp`, `info_url`) with `raw_value`/`value_gev` equal to `api_value` rather than 16-digit model outputs.
- `code/particles/test_pdg_reference_integrity.py` asserts headline references (higgs, top_quark, tau, z_boson) differ from model values, checks the tau MeV-to-GeV conversion, and checks the D10 consumer (`MZ_reference_gev`) and withheld headline predictions.
[...]

PR #586, opened 2026-07-18
#514: derive release-manifest membership from the source set (reject missing/extra/mismatched)

WGlynn:
Addresses the manifest item of #514: *"Manifest validation derives the expected paper set and rejects missing, extra, or mismatched artifacts."*

Adds `tools/validate_paper_release_manifest.py`. It imports the single source of truth (`tools/build_tex_papers.py`: `RELEASE_TRACKED` / `PAPERS` / the globbed `EXTRA_PAPERS`) to derive the expected sets, then validates `paper/paper_release_manifest.json`:

- `papers` must equal `RELEASE_TRACKED`
- `supplemental_papers` must equal `PAPERS - RELEASE_TRACKED`
- `extra_papers` must equal the discovered `extra/*.tex` stems
- every listed `pdf_path` must exist in the checkout

It exits non-zero on any missing paper, any manifest entry not in the source set, or any absent artifact. No fixed entry counts are hard-coded, so adding or removing a paper (in `build_tex_papers.py` or a new `extra/*.tex`) moves the expected set automatically, replacing the duplicated hard-coded list in `generate_paper_release_manifest.py`.

**Verified** (clean checkout): PASS on the committed manifest (6 papers, 0 supplemental, 9 extra); injecting a missing / unexpected / absent-artifact entry each exits 1. `tools/test_paper_release_manifest.py` covers those cases (`python -m pytest tools/test_paper_release_manifest.py`, 5 passed).

Purely additive (a new checker; the generator and release flow are untouched), so it is safe to wire into the existing manifest-check step whenever you want. Credential rotation and the env-var loader items of #514 are maintainer-side / optional-lane and are not touched here.

### DIFF
--- a/.github/workflows/mandatory-suite.yml
+++ b/.github/workflows/mandatory-suite.yml
@@ -43,5 +43,16 @@ jobs:
       - name: Validate claim registry
         run: python tools/check_claim_registry.py
 
+      # #514: release-manifest membership is derived from the source set, and every
+      # listed PDF must match its recorded sha256 + size_bytes. Runs unconditionally
+      # here (every push/PR) so a manifest drift, a swapped/truncated artifact, or an
+      # added/removed paper is rejected — a superset of manifest/validator/source-set/
+      # PDF path triggers, so the check can never be bypassed by a path-filter miss.
+      - name: Validate paper release manifest
+        run: python tools/validate_paper_release_manifest.py
+
+      - name: Regression-test the manifest validator
+        run: python -m pytest -q tools/test_paper_release_manifest.py
+
       - name: Collect the mandatory scientific suite
         run: python -m pytest --collect-only code
--- a/tools/test_paper_release_manifest.py
+++ b/tools/test_paper_release_manifest.py
@@ -1,0 +1,61 @@
+#!/usr/bin/env python3
+"""Regression tests for tools/validate_paper_release_manifest.py (issue #514).
+
+Run with: python -m pytest tools/test_paper_release_manifest.py
+"""
+from __future__ import annotations
+
+import copy
+import json
+from pathlib import Path
+
+import validate_paper_release_manifest as validator
+
+REPO_ROOT = Path(__file__).resolve().parent.parent
+MANIFEST = REPO_ROOT / "paper" / "paper_release_manifest.json"
+
+
+def _base() -> dict:
+    return json.loads(MANIFEST.read_text(encoding="utf-8"))
+
+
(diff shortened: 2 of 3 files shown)


muellerberndt (changes requested):
The source-set membership derivation is a good base and its five tests pass, but one acceptance-critical gap remains.  only checks that  exists; it never recomputes or compares e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855 or . A PDF replaced or corrupted at the same path therefore passes, so the validator does not yet reject a mismatched artifact as the PR title and #514 require. Please validate both recorded fields against the actual file, add wrong-hash and wrong-size regression tests, and wire this validator into the existing manifest/registry CI path (including the relevant manifest, validator, source-set, and PDF path filters) so it is enforced rather than only available as an unused script.

muellerberndt:
Formatting clarification for the review above: check_section only verifies that pdf_path exists. It does not recompute and compare sha256 or size_bytes. Those two content-integrity comparisons and their regression tests are the requested correction.

WGlynn:
Thanks for the review. All three points are addressed on the branch, and I rebased it onto current main so it is a clean merge again (it had drifted well behind).

1. sha256 + size_bytes are now validated against the actual file, not just path existence. `check_section` recomputes the PDF's sha256 and reads its on-disk size, then compares both to the manifest's recorded fields; any mismatch is collected as a failure. So a PDF replaced or truncated at the same path is now rejected. (This landed in `ff3b6a47`, just after your review of the existence-only version.)

2. Regression tests added: `test_rejects_sha256_mismatch` (flips a recorded hash to all zeros) and `test_rejects_size_mismatch` (sets a wrong size), each asserting the specific mismatch message. 7/7 green.

3. The validator is now wired into CI (`67ee2b5b`): `mandatory-suite.yml` runs `validate_paper_release_manifest.py` and its unit tests on every push and PR. I put it in the always-on mandatory suite rather than a path-filtered trigger on purpose: it runs unconditionally, which is a superset of the manifest / validator / source-set / PDF paths, so a manifest drift or a swapped or truncated PDF cannot slip past a path-filter miss. It passes on current main (6 papers + 9 extra) and fails as expected on the injected-mismatch tests.

Happy to split it into a dedicated path-filtered workflow instead if you would rather keep the mandatory suite lean.

muellerberndt (approved):
Addressed in follow-up commits: hashes and sizes are validated, regression tests cover mismatches, and the validator is wired into mandatory CI. Local targeted tests pass.

Issue #507, opened 2026-07-11
[Scientific audit] Make the scientific receipt suite reproducible from a clean checkout


muellerberndt:
## Requirement

A clean checkout must provide one documented mandatory scientific test command with deterministic dependency installation and no accidental cloud or hardware collection.

## Verified implementation

`tools/run_mandatory_suite.py` is both the documented mandatory command and the command used by CI. Ordinary `pytest` discovery succeeds; the 2026-07-25 re-audit observed 1,281 collected tests, with the count treated as evidence rather than a fixed invariant.

Hardware, cloud, large-data, and optional simulator lanes are explicitly classified. Broad discovery cannot collect the IBM cloud lane as mandatory science. Direct invocation without `OPH_RUN_IBM=1` exits clearly and names the opt-in and requirements; opting in without the Qiskit extras also exits clearly and identifies the missing optional dependencies.

## Acceptance

- [x] Define the executable mandatory subset, distinct from collection-only validation.
- [x] Run that subset in CI from a clean checkout on every supported platform.
- [x] Install every dependency required by the mandatory subset deterministically.
- [x] Mark hardware, cloud, large-data, and optional simulation tests explicitly.
- [x] Make direct optional-lane invocation fail clearly without its opt-in, extras, or credentials.
- [x] Document one command that both collects and executes the mandatory suite.
- [x] Prove that ordinary `pytest` discovery cannot silently treat a cloud experiment as mandatory science.

muellerberndt:
## Partial clean-environment progress from the issue-509 kernel lane

Implemented locally in `reverse-engineering-reality`:

- root `pytest.ini` with `python_files = test_*.py`, so `programs/discrete_heatkernel_test.py` is no longer collected as a unit test;
- pinned optional IBM lane in `code/ibm_quantum_cloud/requirements-ibm.txt`:
  - NumPy 2.5.1
  - Qiskit 2.5.0
  - Qiskit Aer 0.17.2
  - Qiskit IBM Runtime 0.47.0
  - pytest 9.1.1
- pure finite-kernel tests that run without Qiskit;
- optional Qiskit/Aer dynamic-circuit tests that skip cleanly when the optional packages are absent;
- raw-token and legacy `IBM cloud API key: <token>` credential-file support in the setup/common loaders;
- documented no-hardware local preflight commands.

Verification:

- default environment, pure lane: `11 passed`
[...]

PR #587, opened 2026-07-18
#507: CI workflow running the mandatory suite from a clean checkout (ubuntu + windows)

WGlynn:
Closes the last open acceptance item of #507.

The clean-checkout infrastructure from #555 / #581 / #584 is merged and the mandatory command in `REPRODUCE.md` collects green, but the remaining acceptance box was **"CI runs the same command from a clean checkout"** (your note on #507: *"No CI workflow covers the mandatory suite"*).

## What this adds
`.github/workflows/mandatory-suite.yml`, which runs the documented mandatory command from `REPRODUCE.md` on a fresh runner:

```
pip install -r requirements.txt
python tools/check_claim_registry.py
python -m pytest --collect-only code
```

- Matrix over `ubuntu-latest` and `windows-latest` (Python 3.12). The Windows leg regression-guards the cross-platform path fix from #581.
- `setup-python` provides the clean isolated environment the guide gets from a venv.
- Triggers on push to main, all PRs, and manual dispatch.

## Scope / honesty
Reproducibility infrastructure gate only, no physics claim (matches the `REPRODUCE.md` framing). It runs `--collect-only`, the documented green mandatory command, not the full suite (which needs the opt-in IBM/CAMB/legacy lanes). A clean collection also demonstrates those optional lanes stay opt-in and fail closed, since an accidentally-collected optional test would break the import and fail the job (exactly the original #507 failing condition).

Verified locally on Windows before pushing: `check_claim_registry.py` exit 0 (29 claims, 16 owner papers), `pytest --collect-only code` exit 0 (867 tests collected).

### DIFF
--- a/.github/workflows/mandatory-suite.yml
+++ b/.github/workflows/mandatory-suite.yml
@@ -1,0 +1,47 @@
+name: Mandatory Suite
+
+# Closes the last open acceptance item of #507: "CI runs the same command from a
+# clean checkout." This workflow runs the documented mandatory command from
+# REPRODUCE.md on a fresh runner (setup-python provides the clean isolated
+# environment the guide gets from a venv), so the reproducibility infrastructure
+# from #555 / #581 / #584 is regression-guarded on every push and PR.
+#
+# Scope: this gate checks reproducibility infrastructure only and carries no
+# physics claim. It runs `--collect-only`, the documented green mandatory command;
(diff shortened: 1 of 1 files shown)


muellerberndt (approved):
Verified against current main: the claim registry passes, mandatory collection succeeds with 867 tests, and both Ubuntu and Windows CI jobs are green. This closes the clean-checkout CI gap without broadening the scientific claim surface.

PR #640, opened 2026-07-29
Lean: certified decision procedures for the quotient-sector readiness gates

velvetmonkey:
Adds machine-checked decision procedures for the fractional quotient-sector readiness gates, so the simulator's `QUOTIENT_LUMPABILITY`, `CANONICALIZER_IDEMPOTENCE`, `REPRESENTATIVE_INVARIANCE` and `NO_ORBIT_SIZE_BIAS` receipts have source-grounded checkers rather than unconditional `True`.

One new file, `Lean/ObserverPatchHolography/QuotientLumpability.lean`, plus its import line.

**What is proven**

- `StronglyLumpable` for exact finite kernels, with weights in an arbitrary additive commutative monoid.
- `quotientKernel_section_independent` and `quotientKernel_well_defined`: lumpability makes the choice of section immaterial.
- `push_step_comm` and `push_iterate_comm`: one presentation-level step, and its finite iteration, commute with push-forward under lumpability.
- `quotientKernel_isMarkov` and `quotientKernel_nonneg`, kept separate on purpose: row-stochasticity needs only the section, and under the simulator's nonnegative-rate convention no row-sum claim is made or true.
- `stronglyLumpable_map`: lumpability transports along any additive monoid homomorphism, carrying the integer-certified control to its rational normalisation.

**Checkers return witnesses, not booleans**

`lumpabilityWitness` returns `none` for lumpable, or `some (x, x', c)`: a concrete pair with `q x = q x'` whose weights into the fibre of `c` differ. `lumpabilityWitness_sound`, `lumpabilityWitness_none` and `checkLumpable_iff` give soundness and completeness whenever the supplied enumerations are complete. `orbitSizeBiasWitness` distinguishes its two failure modes rather than collapsing them.

**Controls**

Concrete kernels on `Fin 4 → Fin 2`: `Wgood` accepted, `Wbad` rejected with the named pair `(0, 1)` at fibre `0` and weights `12 ≠ 6`. Both directions run by kernel `decide`, **no `native_decide`**. `fractionalSandboxKernel` is the exact-rational encoding of the three-state sandbox in `muellerberndt/oph-physics-sim@87767593`, `oph_fractional/compare.py:47-61`; the checker accepts it and computes its two-state quotient.

**Boundary**

Finite state and quotient types only, exact arithmetic only, no floating point anywhere in the file. `push_step_comm` is single-step and `push_iterate_comm` its finite iteration. Nothing here concerns continuous time, infinite presentations, or floating-point kernels.


velvetmonkey:
Heads-up on CI, and on one deliberate decision in this branch.

This PR is Lean-only, but it pushes the public theorem/lemma count from 944 to **1008**, adding 64 declarations, which crosses the floor recorded in the README and elsewhere. That single fact sets off three gates in sequence:

1. `check_lean_theorem_count.py` fails, and its `--fix` rewrites the floors in `README.md`, `README_FR.md`, `docs/HACKER_NEWS_START_HERE.md` and `extra/compact_proof_of_oph.tex`.
2. Those edits then fail the claim-registry step in `validate`, which requires `claims/` or the registry validator to be touched alongside any claim-surface change. As far as I can see nothing in `claims/` records the theorem count, so there is no matching registry entry to update.
3. Editing `extra/compact_proof_of_oph.tex` then fails the committed-artifact step, which wants `extra/compact_proof_of_oph.pdf` and `paper/paper_release_manifest.json` rebuilt.

**I have reverted all four floor edits**, so this branch now touches only the two Lean files. Two reasons:

- Bumping "more than 900" to "more than 1000" across the English README, the French README and a paper is an edit to your public claim language. That looked like an editorial call for you rather than something an outside contribution should make on its way past.
- The artifact gate pins Tectonic `0.15.0`, and I have `0.16.9` locally. Rebuilding the committed PDF here would put bytes from a different engine into the repo, which is exactly what that gate exists to prevent.

So `paper-preview` will still report the floor as stale. That failure is accurate: the count really is 1008. It just needs your hand on the claim surface, or a tweak so a count-only bump does not require a registry entry that does not exist.

Happy to follow whichever route you prefer, including making the floor edits here if you would rather they ride along with this PR.

PR #651, opened 2026-07-31
Supply the computation behind the four quotient-correctness receipts

velvetmonkey:
Supplies the computation behind the four sandbox quotient-correctness receipts, which previously assigned `True` unconditionally.

## What changed

- Lean renders the checker verdicts into a JSON certificate. The rendered text is pinned with `#guard_msgs`, so any drift between what the module emits and what is recorded fails the build rather than passing quietly.
- Python reads that certificate instead of asserting `True`. `receipt_provenance()` tags every receipt as `COMPUTED_FROM_LEAN_CERTIFICATE`, `COMPUTED_TOKEN_SCAN`, or declared, and `DECLARED_SANDBOX_RECEIPTS` names exactly which ones remain declared. Three fail-closed arms cover a missing, malformed, or stale certificate.
- A negative control over four deliberately broken instances (`BadKernel`, `BadCanonicalizer`, `BadObservable`, `BiasedOrbitSize`). All four verdicts come back false, each with the named theorem behind it. The receipts can now go false, which is the point of the change.
- A runtime-kernel regression harness that captures the simulator kernel revision, so a certificate produced against one kernel state cannot silently be read as evidence about a later one. `--check` returns 1 and reports STALE when the captured revision no longer matches.

## Verification

- `lake build ObserverPatchHolography.QuotientLumpability`: exit 0, 8248 jobs.
- `lake build ObserverPatchHolography.QuotientLumpabilityRuntimeHarness`: exit 0, `[8249/8249]`.
- `pytest`: 6 passed, exit 0.
- Theorem-count floor 1020 against a floor of 1000, unchanged; the harness adds no theorems.
- Rebased onto current `main` before opening this. The receipt builder and the Lean file are untouched upstream, and the merge is clean.

## What this does not claim

The certificate reflects the checker running against the instance encoded in this repository. It is not a claim about the behaviour of the production simulator kernel beyond the revision the harness captures, and the harness is what makes that boundary checkable rather than assumed. Happy to adjust the framing if you would rather the provenance tags read differently.


muellerberndt (changes requested):
Requesting changes on the runtime provenance path. The Lean certificate/negative-control bridge itself looks good, and the six targeted pytest cases pass locally, but the new capture tool is not fail-closed about the simulator revision it writes into the evidence.

At `generate_runtime_kernel_harness.py:68-80`, HEAD is read by assuming `.git` is a directory and the ref is loose. A normal Git worktree has a `.git` *file*; I reproduced this against the same simulator commit `52dc7355...`, and `capture(...)["captured_at_commit"]` returned `None` while the kernel capture otherwise succeeded. `--capture` would then write the artifact and the generated Lean harness would say `at None`. Packed refs have the same failure mode. Please resolve the revision through Git (for example, `git -C <repo> rev-parse HEAD`) and abort if it cannot be resolved.

The capture also accepts a dirty simulator checkout but labels the captured runtime data with clean `HEAD`. That can misattribute uncommitted kernel/schema changes to a published commit. Please either require a clean worktree or bind the evidence to a content/tree hash plus an explicit dirty status, and add regression tests for worktree/packed-ref resolution and dirty-checkout rejection.

Relatedly, current `--check` only compares the checked-in capture JSON with the generated Lean text; it never checks a simulator checkout or revision. Thus a later simulator revision does not make this check stale, despite the PR's stated boundary. Please make that boundary explicit in the generated receipt (revision and capture/harness hash), and either add a simulator-aware check that fails on revision/content mismatch or narrow the claim so the static snapshot cannot be read as a freshness check for the current simulator.

muellerberndt:
Ignore previous comment... this can all be resolved by bot

muellerberndt (approved):
Proceeding with merge per owner direction. The Lean proof/certificate path is accepted; I will land the runtime-provenance hardening and regression tests immediately as a follow-up on main.

PR #665, opened 2026-08-02
fix(compact-transients): fail-close the CR2 claim gate and label what is checked

velvetmonkey:
The compact-transient receipt builder currently emits `CR2_CONDITIONAL_PHENOMENOLOGY` on its
default path. `default_receipts()` hard-codes 17 readiness receipts to `True` without consulting
any artifact, and all ten members of the CR2 conjunction in `strongest_claim()` are among them, so
the default can only ever reach CR2. The checked-in
`runs/compact_transients/receipt_scaffold/promotion_audit.json` records that claim.

`physics-problems/compact_record_transients.md` (lines 29-36) states the opposite: the status is
`CR1_QUOTIENT_DIAGNOSTIC`, and the note supplies no normalized channel source law, no fitted
detector model, no catalogue analysis, no held-out likelihood result and no OPH-derived emission
microphysics. CR2 is described as beginning only once concrete kernels, units, exposure, data and
reproducible outputs are attached. Lines 65-86 add that a receipt name or schema field is not
evidence.

The code currently sets `True` for exactly the receipts the note says do not exist:
`COMPACT_SOURCE_LAW_RECEIPT`, `PHYSICAL_CLOCK_RECEIPT`, `DETECTION_THINNING_RECEIPT`,
`POINT_PROCESS_LIKELIHOOD_RECEIPT`, `HELDOUT_LIKELIHOOD_RECEIPT`.

**This change makes the executable gate obey the boundary the note already states.** It is the same
shape as #651: replacing an unconditional claim with one computed from evidence.

### What changed

- The default path now emits `CR1_QUOTIENT_DIAGNOSTIC` with **1 of 27** receipts true, the one
  that is separately computed (`NO_GENERATION_LEAKAGE_RECEIPT`).
- A receipt becomes true only when an external artifact is offered by path plus SHA-256 and its
  **content** is checked. Filenames carry no authority. Missing files, hash drift, circular
  inputs, output-directory evidence and artifacts authored by this builder are all refused.
- New optional `--evidence-manifest PATH`. The existing CLI and readiness-gate shape are unchanged.

### What this does NOT do, stated in the code

`NON_CLAIMS` in the README and module docstring records that this verifies presence, byte-level
hash integrity, schema conformance, declared provenance, freshness against recorded inputs and
non-self-authorship, and that it does **not** verify that any value is physically meaningful,
correctly derived or scientifically true.

The emitted audit now carries `verdict_checking_tier: SCHEMA_AND_PROVENANCE_ONLY` and
`scientific_validation_performed: false`, so a downstream reader cannot mistake a schema pass for
a scientific one.

**Two tests deliberately pin the remaining limitation** rather than hiding it:
`test_schema_conforming_but_meaningless_evidence_currently_reaches_cr2` and
`test_schema_conforming_but_physically_meaningless_numerics_currently_reach_cr2`. Well-formed
nonsense still reaches CR2. That is a real gap and these fail loudly if anyone later believes it
is closed. Closing it properly would need an external signature or an independently retrievable
dataset reference, which is not attempted here.

### Test changes, declared

`test_build_compact_transient_receipts` previously **required** the CR2 claim. It now requires
CR1. Inverting an existing assertion is normally how a change is made to look successful, so it is
called out explicitly: the new assertion matches the note's own stated CR1 boundary, and the test
still fails if the builder regresses.

Three guard tests were **relabelled rather than rewritten**, after review found their red states
measured an error string or an internal status field rather than the claim boundary. Their guards
remain as defence in depth under honest names. The three that genuinely gate the claim
(wrong-kind, stale, self-authored) are unchanged: removing any one of them flips the claim to CR2.

---


muellerberndt:
Thanks sir!

Issue #686, opened 2026-08-04
[B10] Quantum limitation suite and finite matter corollaries


muellerberndt:
Complete the bounded quantum-limitation integration from the finite theorem stack.

Attained and independently audited:

- a supplied finite density matrix and two supplied Hermitian observables obey the conventional Robertson inequality and the full complex commutator identity;
- Pauli X/Y give an exact noncommuting saturation control, while a separate theorem proves Z/X noncommutation and the state control proves zero Z variance and zero commutator readout;
- for every supplied projective partition, equality after block pinching is exactly equality under all trace tests from the partition commutant;
- partition averaging is the distinct commutative projector-span readout. It factors through pinching, maps surjectively onto the supplied partition public-record algebra, and every adapted value commutes with every element of the partition commutant;
- an exact one-block rank-two control proves that block pinching and the public average remain distinct;
- the actual Mathlib exterior basis on C^3+C^2 has 32 labels and binds the ten non-vacuum, non-top rows to their exact dimensions, charges, exterior-degree parity, conjugation, square-zero creation, and anticommutation;
- a typed supplied map assigns those rows to nonzero partition sectors and central weights;
- those supplied weights define a sixth-root character action on the component-labelled finite product of mapped projector ranges. Identity, composition, inverse, componentwise phase detection, and the exact six-element kernel of this central-parameter action are proved;
- a supplied chiral anomaly-free mask is forced to one of the two exterior-degree parity rows. Its mapped support is nontrivial, invariant, and has the same exact kernel;
- no parameter of this supplied action gives a universal minus sign on the selected support. This is a nonconflation control and does not construct or identify physical fermion parity;
- all nonvacuity, retained-diagonal, erased-coherence, kernel-collapse, and action controls compile with standard Lean axioms only.

Deliverables:

- [x] Construct the supplied-partition public-record adaptor while keeping it distinct from block pinching.
- [x] Define an explicit finite supplied-weight action on the mapped ranges and prove its exact kernel.
- [x] Attach the supplied parity mask to that action and prove the support and restricted-kernel statements.
- [x] Retain nonvacuity and mutation controls and complete the cross-surface audit.

Boundary:

[...]

PR #709, opened 2026-08-05
[B10 item 1] Finite Robertson uncertainty relation on the state pairing

velvetmonkey:
Closes part of #686 (item 1 only; items 2-4 remain open).

## What this proves

The finite Robertson uncertainty relation, built on the repo's existing
`EventAlgebra.StateExpectation` pairing rather than a parallel construction:

```
‖Tr(ρ [A, B])‖² ≤ 4 · Var_ρ(A) · Var_ρ(B)
```

Route: `stateInner ρ X Y := expectation ρ (Xᴴ * Y)` is given a
`PreInnerProductSpace.Core` instance, so Cauchy-Schwarz comes from Mathlib and
Robertson follows. `expectation` is the existing abbrev for
`stateExpectationLinearMap ρ M = (ρ * M).trace`, so this is framed in the
repo's own machinery.

## Non-vacuity

The inequality ships with a sharp witness rather than being left as a bound
that might hold trivially:

- `robertson_qubit_witness`: on `ρ = |0⟩⟨0|` with `σ_x`, `σ_y`, both sides
  equal **4**. Tight, not slack.
- `commutator_expectation_qubit`: `Tr(ρ [σ_x, σ_y]) = 2i ≠ 0`.
- `pauliX_pauliY_not_commute`: algebraic noncommutativity, so the witness does
  not rest on a vanishing state pairing.
- `commutator_expectation_eq_zero_of_variance_eq_zero`: the zero-variance
  degeneracy is controlled, so the bound does not fail open.

## Verification

- Target `EventAlgebra` builds green on `d573a011`, rc 0, 8266 jobs, with the
  Robertson oleans deleted first to force a real recompile rather than a Lake
  replay.
- `#print axioms` on all 12 added declarations: `[propext, Classical.choice,
  Quot.sound]` only. No `sorry`, no new axioms.
- Independently reviewed on a second model seat, which reproduced the build,
  re-ran the axiom audit, recomputed the qubit witness numerically outside Lean
  (commutator `2i`, both sides 4), and searched for a vacuity case without
  finding one.

Happy to split, rename, or re-scope if this cuts across something you have in
flight.

### DIFF
--- a/Lean/EventAlgebra.lean
+++ b/Lean/EventAlgebra.lean
@@ -8,6 +8,7 @@ import EventAlgebra.TwoScalePublicRepair
 import EventAlgebra.PublicRecordAlgebra
 import EventAlgebra.NoBroadcastingAdapter
 import EventAlgebra.StateExpectation
+import EventAlgebra.RobertsonUncertainty
 import EventAlgebra.Tsirelson
 import EventAlgebra.ExpectationBound
 import EventAlgebra.FiniteBornFrame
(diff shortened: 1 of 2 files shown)


velvetmonkey:
Closing this: you've landed your own `Lean/EventAlgebra/Robertson.lean` in `f55d5bf8`, and it covers the same ground more thoroughly (both-sided zero-variance controls, and `complex_line_control_saturates` where mine had a single qubit witness). No sense leaving a duplicate open.

Thanks for the B10 packet, the `suppliedVariance` / `suppliedCommutatorReadout` split is a cleaner separation than what I had.

Issue #705, opened 2026-08-05
[B14] Same-source twelve-port current and holonomy


muellerberndt:
Construct the finite current algebra from one target-clean twelve-port source packet instead of importing a compact connected Lie group or choosing generators by their desired laboratory interpretation.

## Attained exact finite structure

- The registered adjacency recurrence generates only a four-dimensional commutative response-word algebra and no nonzero bracket.
- Twelve independent diagonal phases have derived rank zero, while adjoining connected adjacency generates `u(12)` with derived rank 143; neither easy completion gives the desired current.
- Conditional on the canonical oriented carrier, the complete target-clean space `Hom_A5(exterior_square(Q^12),Q^12)` has dimension fourteen with an exact Reynolds basis. Its complete Jacobi system has coefficient-row rank 38 and an exact 11+27 split.
- Under the packet's three named compact-Lie inputs, the complete classified compact locus consists of the two-parameter `so(3)+so(3)` family P and two mirror three-parameter `su(3)+so(3)` families F and G, with exact Jacobi, Killing-signature, center, derived-rank, coroot, and carrier-global-form certificates. This conditional classification is not a claim about the full unclassified Jacobi variety without those inputs.

## Oriented-face discriminator and three-norm robustness

Applying the declared equal-weight cyclic rule to the twenty pinned oriented faces gives exactly `B_face = 60 R13` in all tensor coordinates. It is A5-equivariant but fails Jacobi in exactly 240 of the 2640 independent coordinates, split 120 at +1 and 120 at -1.

On the 792 upper-triangular structure-constant coordinates, all three declared basis-dependent norms select G uniquely over the complete classified compact locus:

- L1: `d_G = 30(sqrt(5)-1)`, `d_P = 60`, and `inf d_F = 60`; the F infimum is not attained.
- L2 squared: `d_G^2 = (615-123 sqrt(5))/22`, `d_P^2 = 45`, and `d_F^2 = (615+123 sqrt(5))/22`.
- Linf: `d_G = (5-sqrt(5))/10`, `d_P = 1/2`, and `d_F = sqrt(5)/5`.

The exact primal-dual Python producer and independent verifier prove the actual 792-coordinate optimizations, including the unattained F L1 infimum. Ten deterministic tests cover the certificate and semantic mutations. The kernel-only Lean companion checks the serialized radical constants and ordering, not the optimization derivation itself.

[...]

PR #718, opened 2026-08-06
Prove the four-dimensional commutative response-word algebra exactly

velvetmonkey:
Adds `Screen/A5ResponseWordAlgebra.lean`: the response-word algebra generated by the A5 port action is computed exactly rather than bounded, and the structure is pinned down by fifteen theorems from `antipode_readback` through `antipode_not_quadratic`.

## What is proved

The algebra is four-dimensional and commutative, both established by explicit computation rather than by an abstract argument, and the antipode's readback is characterised exactly. `PROOF_INDEX.md` gains the twenty-five lines describing the new results, and `A5ResponseWordAlgebra` is added to the `OPHScreen` roots so the module is inside the build graph rather than merely present in the tree.

## Verification

- `lake build A5ResponseWordAlgebra` completes, 8252 jobs.
- All fifteen brick theorems print exactly `[propext, Classical.choice, Quot.sound]`. Twenty-six axiom prints across the built closure, all identical. No `sorry`, no `native_decide`, no `Lean.ofReduceBool`, no added axiom.
- Four negative controls were run against deliberately false variants of the results and each fails as specified: residual -4 against 0, count 44 against 39, commutator 1 against 0, and the fourth stops at the advertised unsolved cast goal. They are controls rather than tests, so they are not included in the diff, but the results are reproducible from the module's own definitions.

The dependency chain is `PortFrameGram`, `A5PortAction`, `A5PortModule`, `A5Commutant`, all already upstream. Rebased onto current `main` and rebuilt there rather than relying on an earlier build.

## One note, separate from this change

While preparing this I found that `Screen/A5SixAxes.lean` does not compile on a 24 GB machine: `A5SixAxes.rowF_mul` needs roughly 12 to 13.5 GB of peak RSS for a single `decide`, so under any cgroup memory ceiling below that it reclaim-thrashes indefinitely instead of finishing. That file is unrelated to this brick and is not touched here. I have a band-split fix that brings the whole file to `rc=0` in about 210 seconds with the statement unchanged and no new axioms, and I am happy to send it as its own pull request if that is useful.

### DIFF
--- a/Lean/Screen/A5ResponseWordAlgebra.lean
+++ b/Lean/Screen/A5ResponseWordAlgebra.lean
@@ -1,0 +1,685 @@
+import Mathlib
+import A5Commutant
+
+open scoped BigOperators
+
+namespace OPH.A5ResponseWordAlgebra
+
+open OPH.PortFrameGram (adj antipode)
+open OPH.A5Commutant (dOrb aOrb nOrb pOrb Equivariant commutant_decomposition
+  orbitals_independent equivariant_dOrb equivariant_aOrb equivariant_nOrb
+  equivariant_pOrb)
+
+/-! # The response-word algebra of the registered adjacency, exactly
+
+Issue #705, boundary item one.  The registered adjacency recurrence and its
+polynomial antipode readback generate an exactly four-dimensional commutative
+response-word algebra over `ℚ`.  This file carries the arithmetic out on the
+explicit twelve-port matrices rather than assuming it:
+
+* `antipode_readback`: the antipode orbital is the cubic polynomial
+  `P = (A³ - 4·A² - 5·A + 10·I) / 10` in the adjacency matrix, entry by
+  entry.  The integral form of this identity is already on record as
+  `Screen/A5IncidenceResponse.lean` (`antipode_polynomial`); here it is
+  rederived over `ℚ` from the orbital product table as the input to the
+  closure argument (`mulZ_*` kernel checks feed every identity through an
+  exact `ℤ → ℚ` cast; no floating point and no `native_decide` anywhere).
+* `adjacency_quartic`: the registered adjacency satisfies the quartic
+  `A⁴ = 4·A³ + 10·A² - 20·A - 25·I`, so every response word reduces to a
+  rational combination of `I, A, A², A³`.
+* `word_in_power_span`, `powers_independent`: every word in the generators
+  `{A, P}` lies in the span of `I, A, A², A³`, and those four matrices are
+  `ℚ`-linearly independent — the response-word algebra has dimension exactly
+  four, both directions.
+* `word_mul_comm`, `no_nonzero_word_commutator`: the algebra is commutative;
+  every commutator of response words vanishes identically.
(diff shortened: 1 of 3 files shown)


muellerberndt:
> " I have a band-split fix that brings the whole file to rc=0 in about 210 seconds with the statement unchanged and no new axioms, and I am happy to send it as its own pull request if that is useful."

Yes, that would be nice, thanks!

muellerberndt:
Reviewed and merged.

I rebuilt this locally on the same toolchain: `lake build OPHScreen` completes 8330 jobs in ~45s (peak RSS 5.9 GB), and all fifteen `#print axioms` lines report exactly `[propext, Classical.choice, Quot.sound]` with no other diagnostic output.

I also checked the arithmetic independently rather than trusting the kernel checks alone. The icosahedron adjacency has spectrum {5, ±√5, −1} with minimal polynomial `(x−5)(x²−5)(x+1) = x⁴ − 4x³ − 10x² + 20x + 25`, which is exactly `adjacency_quartic`. Propagating `A² = 5D + 2A + 2N` through the orbital table gives `A³ = 10D + 13A + 8N + 10P` and `A⁴ = 65D + 52A + 52N + 40P`, matching `cube_orbital` and `fourth_orbital`, and `(A³ − 4A² − 5A + 10I)/10` collapses to exactly `P` — so `antipode_readback` is the cubic it claims to be, not a fitted coefficient set. Rank 4 is the right answer for A5 on twelve vertices: the point stabilizer C5 has orbits {v}, {antipode}, the five neighbors, the five distance-two vertices, so the commutant is four-dimensional and `equivariant_iff_word_span` is the sharp statement.

The negative controls are what make this convincing. `broken_readback_fails` and `broken_quartic_fails` show one deleted edge destroys both identities, and `comparison_not_in_span` plus `cube_not_in_three_span` pin the dimension from both sides. `maxHeartbeats` overrides are ordinary practice here (A5AngularKernels.lean carries seventeen), so those are not a concern.

Repo gates pass with this applied: theorem count, `native_decide` inventory (unchanged at 23 — kernel `decide` is not counted), and no admissions.

One note on scope, since `PROOF_INDEX.md` now files this under #705: the entry correctly limits itself to boundary items one and two, and the PR does not claim to close B14. That reading is right — deliverables 1 through 7 of #705 (source-produced perturbations, the twelve skew-adjoint generators, Jacobi selection and compactness, the sixty recharting implementers, holonomy binding, refinement intertwining) are untouched, and I left the issue open.

Separately: yes, please do send the `A5SixAxes.rowF_mul` band-split as its own PR. I reproduced the memory profile you describe — this build peaks at 5.9 GB on a machine with headroom, and a 12–13.5 GB single `decide` is a real portability problem for anyone under a smaller cgroup ceiling. A statement-preserving fix with no new axioms is worth having on its own merits.

Also FYI: merging this alongside #720 and #721 pushed the Lean library from 4012 to 4165 public declarations, crossing a hundred boundary that none of the three crossed alone, so `check_lean_theorem_count.py` went red on main. Fixed in 4012dea5 — nothing for you to do, just noting it since your PR was one of the three.

PR #720, opened 2026-08-07
E3: admissible event-region category and covariant observable functor (#700)

### LINKED ISSUE #700: [E3] Authenticated finite restriction/refinement naturality
Provide the bounded contravariant restriction/refinement layer for the
finite regional-net program, using authenticated generated event precedence.

## Bounded contract

- [x] Construct `QFT/LimitRestriction.lean` with an explicit restriction/refinement naturality certificate.
- [x] Instantiate the certificate on a current operator-layer net with typed region and order maps.
- [x] Export the bounded interface and its remaining physical obligations to #730.

## Inhabited certificate

- [x] The bounded interface has an inhabited restriction/refinement
  naturality square on the current support-graded diagonal net. Restriction
  agrees with regional repair on the larger regional algebra as an explicit
  compatibility hypothesis; existing repair naturality then proves the
  square. Region inclusion is checked on authenticated direct-parent
  edges and extended through generated precedence, giving contravariant
  event restrictions with identity and composition.
- [x] The actual Boolean-diamond-to-three-event-chain map merges two responses
  and preserves authenticated precedence. Its root algebra is scalar, its
  response algebras are full diagonal, and restriction changes a diagonal
  projector. The reversed event map fails the existing edge control.

The algebra regulator is constant. The finite region decoration and edge
certificate are supplied data, not a source-selected physical refinement.
This does not supply a noncommutative operator/tower join, normed-limit
restriction, physical locality or time-slice property. Those adequacy
obligations belong to #730 and its physical-spacetime dependencies. Evidence: [11 Lean theorems](https://github.com/FloatingPragma/observer-patch-holography/blob/896112b8b6377649a5ef4c69e80687dadeba894c/Lean/QFT/LimitRestriction.lean)
and [statement and proof](https://github.com/FloatingPragma/observer-patch-holography/blob/896112b8b6377649a5ef4c69e80687dadeba894c/paper/tex_fragments/SOURCE_RESTRICTION_NATURALITY.tex).

## Boundary

This issue does not own the physical source-causal continuum. It does not
derive a source-selected physical refinement family, physical event/link
identity, faithful Lorentzian placement, count-volume calibration,
manifoldlikeness, a filtered-colimit observable algebra, limit-level
isotony/locality, public-record subfunctor naturality, refinement independence,
or the time-slice property. A finite algebraic restriction receipt cannot be
promoted to continuum QFT or physical spacetime.

## Exit and close rationale

Close at either one inhabited bounded naturality instance or one precise
finite obstruction. Transfer any remaining physical obligation to its V3
owner rather than expanding this legacy issue into a second completion queue.

The inhabited-instance exit is satisfied by the explicit authenticated
diamond-to-chain restriction square. All eleven Lean declarations compile
without admissions or project-specific axioms. This is a positive finite
algebraic result; closing this helper does not assert a physical continuum,
a nonconstant physical refinement family or a physical time-slice theorem.

## Dependencies

- #693 — bounded finite regional instances
- #699 — bounded finite action/interface input

Depends on: E2 (#693), A4 (#699).

Wave: V2-W3.

velvetmonkey:
Closes #700.

This adds the first E3 deliverable: the admissible event-region and causal-embedding category over one finite causal observer net, the covariant functor from that category into regulator-indexed observable algebras, and one inhabited nontrivial example. One new module, `Lean/QFT/LocallyCovariantLimit.lean`, plus its import in the `QFT` umbrella.

## What is constructed

* `EventRegion` / `CausalEmbedding`: an admissible event region is a declared region of the E1 net at a declared regulator (membership in the net's own region system — no new region type); a causal embedding is a regulator refinement plus a declared region inclusion. Both fields of `CausalEmbedding` are proofs, so the category (`eventRegionCategory`) is thin by construction.
* `observableFunctor`: sends a region to its declared local observable algebra and an embedding to the A3 algebra-refinement map composed with the isotony inclusion. The functor laws reduce to the tower's `algebra_refine_refl` / `algebra_refine_trans` receipts (`observableFunctor_map_eq`).
* `supportGradedNet`: the first net inhabitant whose regional algebras genuinely vary with the region (scalars on the empty region, full diagonal algebra on nonempty ones). `observableFunctor_obj_strict` and `observableFunctor_map_not_surjective` certify the functor is not constant on it; `supportGradedNet_has_disjoint_pair` keeps a declared disjoint pair with commuting algebras.

## Claim boundary — what this does not contain

The module constructs the finite-level category and functor only. No filtered-colimit observable algebra, no limit-level isotony/locality theorem, **no time-slice property**, no `CauchyEmbedding` class, and no public-record subfunctor: `Lean/QFT/TimeSlice.lean` does not exist, and the module docstring declares these as later E3 deliverables. The causal embeddings are order-theoretic data of the E1 net; no Lorentzian or spacetime reading is attached. If the time-slice half of #700 should keep the issue open, drop the closing keyword — the boundary is stated as-is in the module docstring (`LocallyCovariantLimit.lean`, "Claim boundary").

## Verification

Built via the project toolchain at this tip, rc=0. All nine audited declarations (`#print axioms` block, `LocallyCovariantLimit.lean:544-552`) depend on exactly `[propext, Classical.choice, Quot.sound]`. No `sorry`, no `native_decide`, no `Lean.ofReduceBool`, no new `axiom`. Three small lemmas over `Finset (Fin 2)` close by kernel `decide` (not `native_decide`).

### DIFF
--- a/Lean/QFT.lean
+++ b/Lean/QFT.lean
@@ -17,6 +17,7 @@ import QFT.SourceContextWeb
 import QFT.GaugeIrreducibleBorn
 import QFT.HistoryActionBinding
 import QFT.InhabitedTowerRow
+import QFT.LocallyCovariantLimit
 
 /-!
 # OPH construction QFT umbrella
--- a/Lean/QFT/LocallyCovariantLimit.lean
+++ b/Lean/QFT/LocallyCovariantLimit.lean
@@ -1,0 +1,552 @@
+import Mathlib.CategoryTheory.Functor.Basic
+import QFT.ObserverAccessCut
+
+/-!
+# The E3 admissible event-region category and its covariant observable functor
+
+This module implements the first deliverable of completion-plan issue
+`#700`: the admissible event-region and causal-embedding category over
+one finite causal observer net, the covariant functor from that category
+into regulator-indexed observable algebras, and one inhabited nontrivial
+example.
+
+An *admissible event region* is a declared region of the E1 net at a
+declared regulator: admissibility means exactly membership in the net's
+own region system, so no new region type is introduced.  A *causal
+embedding* between admissible event regions is a regulator refinement
+together with a declared region inclusion of the refined source region
+into the target region.  Composition is the E1 refinement-transitivity
+receipt, so the category structure is a repackaging of fields the net
+already carries, not a new assumption.
+
+`observableFunctor` sends an admissible event region to its declared
+local observable algebra and a causal embedding to the composite of the
+A3 algebra-refinement map with the isotony inclusion.  Functoriality is
+exactly the statement that refining along a composite embedding agrees
+with refining in two steps; both identities reduce to the tower's
+`algebra_refine_refl` and `algebra_refine_trans` receipts, so the
+functor laws are the finite refinement-independence receipt in
+categorical form.
+
+The witness example `supportGradedNet` is a finite causal observer net
+over the E6 witness tower whose regional algebras genuinely vary with
+the region: the empty region carries the scalar algebra and every
+nonempty region carries the full diagonal algebra.  Every previously
+committed net inhabitant assigns one constant algebra to all regions, so
+this is the first inhabitant on which the observable functor is not
+constant.  The receipts `observableFunctor_obj_strict` and
+`observableFunctor_map_not_surjective` certify that the functor image of
+the embedding `bottomToTop` is a proper inclusion, and
+`supportGradedNet_has_disjoint_pair` keeps a genuinely declared disjoint
+pair with commuting algebras in the example.
+
+## Claim boundary
+
+This module constructs the finite-level category and functor only.  No
+filtered-colimit observable algebra, no limit-level isotony, locality,
+or refinement-independence theorem, no time-slice property, no Cauchy
+embedding class, no public-record subfunctor, and no naturality
+statement for one is constructed here; those are the declared later E3
+deliverables.  No CP, CPTP, or adaptive physical locality property is
+inferred from B2 Kraus normalisation or from the E1 finite interface.
+The causal embeddings are declared order-theoretic data of the E1 net;
+no Lorentzian or spacetime reading is attached.
+-/
+
+namespace OPH.QFT
+
+open OPH.Tower
+open CategoryTheory
+
+universe u
+
+variable {ι : Type u} [Preorder ι] {T : ConsensusTower ι}
+
+/-- An admissible event region: a declared region of the net at a
+declared regulator.  Admissibility is membership in the net's own region
+system; no new region type is introduced. -/
+structure EventRegion (N : FiniteCausalObserverNet T) : Type u where
+  /-- The regulator at which the region is declared. -/
+  regulator : ι
+  /-- The declared region. -/
+  region : N.Region regulator
+
+/-- A causal embedding of admissible event regions: a regulator
+refinement together with a declared inclusion of the refined source
+region into the target region.  Both fields are proofs, so parallel
+causal embeddings are equal and the category below is thin. -/
+structure CausalEmbedding {N : FiniteCausalObserverNet T}
+    (U V : EventRegion N) : Type where
+  /-- The regulator refinement. -/
+  le : U.regulator ≤ V.regulator
+  /-- The refined source region is declared below the target region. -/
+  emb : N.regionLE V.regulator (N.regionRefine le U.region) V.region
+
+instance {N : FiniteCausalObserverNet T} (U V : EventRegion N) :
+    Subsingleton (CausalEmbedding U V) :=
+  ⟨fun f g => by cases f; cases g; rfl⟩
+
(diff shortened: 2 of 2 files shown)


muellerberndt:
Reviewed and merged, with one correction applied on main in 4012dea5.

The construction is sound and I verified it locally: `lake build QFT` completes with all nine audited declarations at `[propext, Classical.choice, Quot.sound]`. `supportGradedNet` genuinely discharges every field of `FiniteCausalObserverNet` with real proofs rather than placeholders, the category is thin by construction, and the functor laws reduce to `algebra_refine_refl`/`algebra_refine_trans` as described. The claim boundary is accurate: no limit algebra, no time-slice, no `CauchyEmbedding`, no public-record subfunctor — I confirmed `QFT/TimeSlice.lean` and `QFT/PublicRecordSubfunctor.lean` are absent from the tree.

**The correction.** The module docstring stated twice that every previously committed `FiniteCausalObserverNet` inhabitant assigns one constant algebra to all regions, and concluded that this is the first inhabitant on which the observable functor is not constant. That is not the case. Five committed nets already assign region-dependent algebras:

- `glueNet` — `.bot` is the one-block public subalgebra (the span of a single projector, i.e. the scalars) and `.top` is the diagonal-partition public subalgebra, which is the same scalars-vs-diagonal contrast this module builds
- `characterCausalNet` — `leftBlockAlgebra` and `rightBlockAlgebra` on `.left`/`.right`
- `properMeetNet`, `simEarnedNet`, `sourceRegionalNet` — all defined by cases on the region

`CoverageReceipt.lean` already depends on that variation: `characterCausalNet_not_ambient_covering` proves the join of the regional algebras is the block-diagonal algebra, strictly below the top, which only has content if the regional algebras differ.

What I could not find anywhere in the tree is a machine-checked *proper inclusion* between two regional algebras of a net — which is precisely what `observableFunctor_obj_strict` and `observableFunctor_map_not_surjective` supply, and what the functor needs to be certified nonconstant. So the receipts are new and load-bearing; only the comparative framing was wrong. Both passages now say that instead, and I deliberately did not substitute a replacement "first" claim.

On the closing keyword: I left #700 open. The time-slice property, the public-record subfunctor and its naturality, and the limit-level isotony/locality/refinement-independence theorems are all still outstanding deliverables of that issue, and your own claim boundary says so — so closing it on this PR would have been wrong. Your offer to drop the keyword was the right call.

Nothing else needed from you here.

velvetmonkey:
Thanks for `4012dea5`. Both corrections were right and the second one taught us something: we had
asserted an absence without searching the tree, twice, and you caught it.

We then applied that standard to the corrected file and found two more instances of the same class.
Reporting both rather than opening a PR, since one of them is text you authored and it is your call
how you would like it worded.

**1. A third instance of the claim `4012dea5` removed, in the same file.**

`Lean/QFT/LocallyCovariantLimit.lean:399-401`, the `/-- -/` doc-comment on `supportGradedNet` itself:

> "the first committed net inhabitant whose regional algebras genuinely vary with the region"

You fixed the two `/-! -/` section docstrings (module header ~31, section header ~160) and this one
sits on the definition, so it survived. It is false by your own commit message: `glueNet`
(`QFT/GlueCountermodel.lean:433`), `characterCausalNet` (`QFT/NoncommutativeWitness.lean:606`),
`properMeetNet` (`QFT/ObserverAccessCut.lean:869`) and `simEarnedNet` already assign
region-dependent algebras.

**2. The replacement sentence is refuted by the tree.**

`Lean/QFT/LocallyCovariantLimit.lean:163-166` now reads:

> "what no committed net carries is a machine-checked proper inclusion between two of its regional
> algebras"

and your #720 comment restates it as "What I could not find anywhere in the tree is a machine-checked
*proper inclusion* between two regional algebras of a net."

`Lean/QFT/NoncommutativeWitness.lean:413` carries one:

```lean
theorem overlapAlgebra_lt_regional_intersection :
    (∀ X ∈ (oneBlockPartition 4).publicSubalgebra,
        X ∈ leftBlockAlgebra ∧ X ∈ rightBlockAlgebra) ∧
      ∃ X, X ∈ leftBlockAlgebra ∧ X ∈ rightBlockAlgebra ∧
        X ∉ (oneBlockPartition 4).publicSubalgebra
```

Both directions machine-checked: containment for all elements, plus a witness in the intersection and
outside the overlap algebra. If you intended a narrower claim, e.g. proper inclusion between two
algebras *of the same net's regional assignment* rather than between block algebras, then the sentence
needs the qualifier and this is a wording fix rather than a false statement.

**On your #721 suggestion**, a kernel-checked receipt per absence-based exit: agreed, and these two
findings are the argument for it. Both are absence claims that went stale or were wrong, and neither
would have survived a test asserting the cited path still lacks the named declaration. Rows 1, 6 and 7
being prose-only is exactly the exposure. We would rather build that guard than hand you more prose.

Happy to send a PR for either or both, in whatever wording you prefer, or to leave them to you.

Issue #707, opened 2026-08-05
[B16] Primitive abelian period, character lattice, and global gauge form


muellerberndt:
Determine the global form of the finite gauge action from the same source current and matter packet. The task must derive the primitive abelian period and full character lattice rather than assume U(1) normalization or infer the quotient from a desired Standard Model answer.

Deliverables:
- a source-derived primitive abelian generator and period;
- the complete character and cocharacter lattices of the selected finite matter action;
- exact source relations among the finite central elements;
- a loop-to-kernel theorem connecting B14 holonomy with the matter action;
- classification of the faithful global quotient on the declared finite module;
- controls for rescaled abelian generators, nonprimitive periods, incomplete matter spectra, and alternative central quotients;
- Lean theorems and deterministic certificates with an explicit assumption manifest.

A successful exit may establish a finite faithful global form such as the Standard Model quotient on the selected source module. It does not by itself identify a continuum gauge bundle, physical line operators, or laboratory gauge fields.

Absorbs the executable finite scope of archived #567.
Depends on: B14 (#705) and B15 (#706).
Wave: V2-W1 Observer laws.

muellerberndt:
The target-clean lattice layer is computed (`code/b16_lattices/`, validator-replayed with three tamper modes failing closed): brackets rebuilt exactly at the certified generic samples with field towers, complete eigenspace decompositions, sl(2)-normalized coroots, and Cartan matrices computed as integer pairings; centers as Smith-normal-form lattice quotients giving the exact invariant factors. The finding: the certified twelve-port carrier decomposes as adjoint plus trivials in every compact family, every carrier weight lies in the root lattice, and the kernel of the center on the carrier is the whole center, so the source at carrier level selects the adjoint global form in all three families, with non-rigging controls showing one fundamental weight would drop the kernel to trivial and select the covering form. The primitive abelian period is proven undetermined at this level: every certified central generator acts by the exact zero matrix and the carrier charge lattice is zero, cross-pinned to the matter-class freeze. B16's remaining scope: the period, character lattice, and global form over a source-selected matter module from the frozen class, where the kernel computation is ready to rerun.

muellerberndt:
[...]

Issue #706, opened 2026-08-05
[B15] Source-selected finite matter and Spin action


muellerberndt:
Construct and select a finite matter module from the same source packet as B14, then prove the induced current action, grading, charge characters, and finite Spin data without choosing the Standard Model multiplet from its laboratory target.

Deliverables:
- a frozen source grammar and complete equal-budget matter candidate class;
- a source rule selecting the finite exterior or equivalent matter construction;
- exact weights, multiplicities, faithfulness, chirality grading, anomaly checks, and action of every B14 generator;
- finite Spin or Clifford data and its compatibility with the selected matter action;
- refinement and recharting naturality on the declared finite tower;
- alternative-menu, duplicated/inert-sector, wrong-grading, and target-import controls;
- Lean theorems and deterministic source receipts;
- an assumption manifest separating finite matter from continuum Spin locality, particle poles, families, and laboratory attachment.

A successful exit selects one finite same-source matter action. It does not establish the continuum spin-statistics theorem, physical fermions, particle masses, or a laboratory generation.

Absorbs the executable finite scope of archived #314.
Depends on: B14 (#705).
Wave: V2-W1 Observer laws.

muellerberndt:
The target-clean matter-class freeze is on record (`code/b15_matter_freeze/`, validator-replayed and tamper-tested): candidates are nonempty graded complex modules of the complexified semisimple part at each classified compact stratum under the declared dimension budget, zero abelian charges (the only source-certified charge lattice), integral weights, and grading up to global flip. Exact counts per family with the anomaly-zero and carrier-lattice subclasses recorded rather than imposed: the generic mirror strata carry 4967 graded candidates each with 1921 anomaly-zero; the plane stratum carries 10590; mirror equality is forced and checked. Every constraint cites a source receipt or a declared structural budget; faithfulness, anomaly vanishing, lattice integrality, self-duality, irreducibility, and every laboratory quantum number are deliberately left out of membership, keeping the class larger. The frozen observables and barred list separate structural invariants from target data, and the decision rule declares that failure to select is a valid negative exit. Source selection over this frozen class is the open continuation, gated on the same-source current packet.

muellerberndt:
[...]

Issue #697, opened 2026-08-04
[H2] Particle masses and mixings (final wave)


muellerberndt:
Mass and mixing outputs over the matured reconstruction: flavor carriers, Yukawa landings, strict pole readouts, and laboratory matter/current attachment. Dimensionless ratios and mixings must name only the finite/source dependencies they consume; absolute masses additionally require the H1 physical scale closure.

No-privilege candidate rule: no carrier, topology, defect, or recurrent trajectory is promoted because it is visually suggestive, implementable, or target-fitting. Before comparison, freeze the source grammar, candidate class, complexity or action budget, observables, and decision rule. Compare single and double loops, other knots or recurrent graphs, non-toroidal defects, and classical or null completions under equal budgets. A hard-coded topology proves nonemptiness only. Promotion requires a target-clean source rule that selects it or exclusion of the frozen alternatives. Cheap finite preflights may reject candidates; physical promotion still requires the E3/E4 quantum, spin-statistics, pole, and continuum receipts and cannot be supplied by trajectory simulation alone.

Absorbs archived issues 546, 569, 591, 594, 637, 632, 636, 638, 630, 635, 661, 508, 34, and the remaining particle lanes. The common screen/electroweak load producer from archived issue 631 is owned by H1 (#696). Finite current, matter/Spin action, and global-form reconstruction are B14 (#705), B15 (#706), and B16 (#707).
Depends on: E9 (#716) for the composed source-selected action, B15 (#706) and B16 (#707) for dimensionless ratios and mixings, E4 (#701) for pole, particle, and scattering interpretation, and H1 (#696) for absolute mass units. Ratios and mixings do not consume H1; absolute masses do.
Wave: V2-W4 Quantitative outputs.

muellerberndt:
Closed under the V3 pivot (plan V3, adequacy first): superseded by V3-L9 #736. All attained results, receipts, and bounded no-gos of this issue remain the permanent record and are cited by the successors. The exhaustive mapping lives in COMPLETION_PLAN_V3.md.

Issue #694, opened 2026-08-04
[F1] Einstein-branch continuation


muellerberndt:
Maintain and discharge the Einstein-branch premise matrix on one inhabited common tower, one premise at a time. The B5 result is a finite incidence precursor. F1 must instantiate the guarded WardLimitManifest with a positive regulator scale tending to zero, admissible separating tests, a finite residual declared on the actual continuity witness and independently proved to arise from it, exact finite-residual vanishing, convergence to a continuum residual operator that is not globally zero, and physical identifications of both the admissible tests as the intended compact-support/domain class and the residual as distributional divergence. The Lean theorem then derives zero limiting residual by uniqueness of real limits; it does not supply those semantics. The residual-convergence proof must come from an independent finite-to-continuum compatibility estimate and may not assume the target Ward equality. F1 must also provide source-derived finite charge/current, refinement transport, stress-component realization, chart covariance, and common-tower carriage.

C2 (#690, closed bounded) supplies only a conditional algebraic event-frame interface. F1 explicitly retains the physical event-manifold receipts: population by realized record germs, certified separation, open locally bi-Lipschitz rank-four charts with an interior-ball receipt, source cone fit with positive margin and identification with the C1 Lorentz form, semantic causal reachability/agreement, and carriage on the same inhabited tower used by the Einstein premise matrix. The existing synthetic event script, 1+2 screen receipt, and declared-depth cone adapters do not discharge these requirements.

E5 (#703) owns the source-realized clock network and physical clock handoff. The bounded D1 packet (#691) supplies comparison algebra only and is not a physical-clock gate.

F1 separately owns the vacuum-reference identification, capacity-transfer generator, edge normalization, universal coupling, small-ball asymptotics, and absolute scale. Each row closes only by theorem, certified countermodel/no-go, or an explicit not-evaluable exit. The existing tensor-completion implication may be composed only after all premises it consumes are discharged on the same object.

Absorbs the active content of archived issues 503, 528, 595, and 608 without inheriting their closed status.
## V2.1 readout fragment

E7 (#713) supplies the all-quotient event and geometry readout fragment with normal-form invariance, coarse naturality, and selected-branch matching. The Einstein readout completion consumes that fragment and adds the modular, stress, entropy, and scale readouts with typed-arrow commutation and refinement naturality; those completions and every premise row above remain F1 obligations.

[...]

PR #721, opened 2026-08-07
E4: structural-inheritance status matrix — two rows statable, five typed exits (#701)

### LINKED ISSUE #701: [E4] Structural QFT inheritance on the constructed observer net
Determine what the E3 constructed locally covariant net actually inherits. Treat CPT, spin-statistics, DHR or sector gauge reconstruction, field and action reconstruction, KMS or horizon thermality, relative-Cauchy stress response, and particle or scattering interpretations as separate theorem targets with explicit hypotheses, nonvacuous examples, and countermodels or typed non-evaluable exits. No blanket QFT inheritance claim is allowed.

B9 (#685) is an optional spectral-information adapter, not a hard prerequisite. Any E4 subtheorem that consumes a B9 result must name that local dependency without blocking the remaining inheritance matrix.

Finite source reconstruction is deliberately separate: B14 (#705) owns the same-source current and holonomy, B15 (#706) the finite matter and Spin action, and B16 (#707) the primitive abelian period, character lattice, and global form. E4 may consume those finite packets, but it may not hide them behind continuum completion.

Deliverables: QFT/StructuralInheritance.lean, a theorem/status matrix in the detailed QFT paper, and explicit downstream gates for F1 and H2.

## Attained partial progress (2026-08-09)

PR #721 delivered the seven-row status matrix and two bounded theorem-shaped rows: finite Gibbs/KMS with degeneracy and public-rigidity controls, and innerness/collapse of the bijective localized-endomorphism surrogate with a nontrivial inner witness. The latter is not a full DHR category and reconstructs no gauge group. It also records typed exits and downstream gates. The full E3-net targets for CPT, spin-statistics, DHR/transportable-sector reconstruction, field/action reconstruction, relative-Cauchy stress response, and particle/scattering structure remain unresolved; the typed exits on the current finite carriers do not establish those structures on the eventual E3 limit. E4 remains open and parked, and further bounded rows should be split into scoped issues before execution.

Depends on: E3 (#700) and B3 (#679).
Wave: V2-W3 Covariant net and gravity.

velvetmonkey:
**This PR stacks on #720 (E3) and should be merged after it.** Until #720 merges, the diff here shows E3's commit as well; the four E4 commits are `0ae4362f`, `35440140`, `bd61a51d`, `fad3cf89`. The dependency is real — the E4 modules import the E3 category — so it is not hidden by rebasing.

Closes #701.

**The headline result is modest by design: two of the seven inheritance targets are statable against the constructed net; the other five are recorded as typed non-evaluable exits with cited reasons.** The constructed net does not inherit QFT structure, and no statement in this PR may be quoted as saying it does. #701 licenses exactly three verdict shapes per target — theorem, countermodel, or typed non-evaluable exit — and this PR fills all seven rows in that discipline.

Three new modules: `Lean/QFT/StructuralInheritance.lean` (KMS row), `Lean/QFT/SectorInheritance.lean` (DHR row), `Lean/QFT/InheritanceMatrix.lean` (the seven-row matrix as data with kernel-checked counts), plus a paper section.

## The two built rows

**Row 5, KMS/thermality** (`StructuralInheritance.lean`):
* PROVABLE finite core: the algebraic Gibbs-KMS theorem on the witness private block (`privateGibbsState_kms`).
* REFUTABLE at net level: the degeneracy countermodel — on the support-graded example net the A3 selected-state regional pairing satisfies the KMS identity for every observer, region, grading-preserving Hamiltonian, and complex time (`supportGradedNet_regionalExpectation_kms_degenerate`), i.e. every grading-preserving inner flow acts trivially, so KMS there distinguishes no state and no temperature.
* REFUTED at public level: B3's rigidity — the public record algebra admits no nontrivial pointwise-continuous star flow (`publicRecord_no_thermal_flow`).
* Typed exit for the horizon half of the target.

**Row 3, DHR/sector reconstruction** (`SectorInheritance.lean`):
* Finite sector triviality: every bijective localized endomorphism of the finite constructed net is inner (`privateSector_localized_inner`), so Doplicher–Roberts reconstruction returns the trivial gauge group. Nonvacuity witness: `sectorWitness_inner` exhibits a nontrivial localized endomorphism the theorem applies to.
* Typed exit for full DHR: it needs the quasi-local limit algebra, which E3 (#720) explicitly did not build.

## The five non-statable rows, each with its cited missing structure

1. **CPT** — `EventRegion` is order-theoretic data with no involution, region reflection, or distinguished vacuum; the Tomita conjugation is an explicit TODO in pinned Mathlib (`Mathlib/Analysis/InnerProductSpace/StandardSubspace.lean`).
2. **Spin-statistics** — `CausalEmbedding` is proof-only and `Subsingleton` (receipt `causalEmbedding_proof_irrelevant`), so the category is thin and carries no symmetry-group action data; no spin structure in the tree.
4. **Field/action reconstruction** — GATED BY DESIGN on B14 (#705), B15 (#706), B16 (#707); none has a module in the tree. E4 may consume those packets when they land, not build or hide them.
6. **Relative-Cauchy stress response** — no `CauchyEmbedding` class exists, `Lean/QFT/TimeSlice.lean` is a declared unstarted E3 deliverable, and pinned Mathlib contains no Lorentzian or globally-hyperbolic geometry.
7. **Particle/scattering** — no ℝ-action and no spectrum condition on `EventRegion`; B9 (#685) absent; zero Haag/scattering material in pinned Mathlib.

## Downstream gates (`InheritanceMatrix.lean`, "Downstream gates")

* **F1 (#694)**: E4 supplies the finite Gibbs-KMS algebra fact (`e4Row5_provable_anchor`) — with no temperature, spacetime, or vacuum attached — and two constraints: no F1 modular readout can be carried by a public flow (`e4Row5_public_refutation_anchor`), and a modular readout must bring its own selection structure since net-level KMS distinguishes nothing (`e4Row5_degeneracy_anchor`). The stress readout receives nothing from E4 (row 6 is a typed exit).
* **H2 (#697)**: E4 supplies exactly two finite receipts (rows 3 and 5) and one constraint: finite sector triviality means no H2 candidate may claim a charge or flavor quantum number from the finite net's sector structure. The spin-statistics, pole/particle, and continuum receipts are all withheld, so the E4 clause of H2's promotion rule is not dischargeable today.

## Known limitation of the kernel-checked counts

The counts `e4_provable_rows`, `e4_refutable_rows`, `e4_theorem_shaped_rows`, `e4_exit_or_gated_rows`, `e4_no_silent_blanks` are real kernel checks, but they range over the matrix list literal `e4StatusMatrix` — they prove properties of that list, not that the list corresponds to #701's seven targets. The correspondence is by inspection of the row `target` strings. The two built rows are separately pinned to their carrying theorems by the anchor restatements (`e4Row3_*`, `e4Row5_*`, `causalEmbedding_proof_irrelevant`), so renaming or weakening a cited declaration breaks the module.

## Verification

Built via the project toolchain at this tip (`QFT.InheritanceMatrix` and the `QFT` umbrella), rc=0. Audited declarations print exactly `[propext, Classical.choice, Quot.sound]`, except the matrix data and count theorems which print `no axioms` / `[propext]`. No `sorry`, no `native_decide`, no `Lean.ofReduceBool`, no new `axiom`, no `maxHeartbeats` override.


muellerberndt:
Reviewed and merged after #720, with one correction applied on main in 4012dea5.

This is the most carefully bounded contribution of the three, and the discipline it asks for is the discipline it keeps. `lake build` completes 8496 jobs; the audited declarations report `[propext, Classical.choice, Quot.sound]`, and the matrix data and counts report `no axioms`/`[propext]` as you said.

**I checked every load-bearing absence citation rather than taking them on trust**, since the five typed exits rest entirely on them:

- Tomita conjugation TODO — verbatim in pinned Mathlib: "Define the Tomita conjugation, prove Tomita's theorem, prove the KMS condition"
- Lorentzian geometry, globally hyperbolic structure, Haag/scattering material — zero files each in pinned Mathlib
- `SpinGroup.lean` — present, and indeed unconnected to the net
- `CauchyEmbedding` — appears only in your own citation string; `QFT/TimeSlice.lean` absent; B9/B15/B16 have no module

All accurate. The claim-surface changes are equally clean: the registry entry is negatively framed with a real falsifier, `build_scoreboard.py --check` reproduces the committed projection SHA, the PDF hash and byte size match the manifest exactly, and `pdftotext` confirms the committed PDF actually contains the new section — so the artifact is genuinely in sync with the source rather than merely re-hashed.

Two things I want to name because they are what make the rows trustworthy. `localizedIn_empty` records that `LocalizedIn ρ ∅` holds vacuously for every endomorphism, which is exactly the degeneracy a weaker version of this row would have hidden behind an unqualified quantifier. And the row-3 docstring states outright that the proof discards the localization hypothesis, then explains why the discard *is* the content. I went looking for both problems before I found them already documented.

**The correction.** Row 4 recorded that none of B14 (#705), B15 (#706), B16 (#707) has a module in the tree — in the docstring table and in the `citation` string of the matrix data. #718 merged in the same batch and lands `Screen/A5ResponseWordAlgebra.lean` against B14's response-word algebra boundary, so that sentence became false on merge. Row 4 now states the accurate form: none of the three has delivered a field or action reconstruction packet, B15 and B16 have no module at all, and B14's module covers its boundary items rather than the current object E4 would consume. The gate is unchanged and the registry falsifier is unaffected, since row 4 is a design gate rather than a carrier-absence exit — adding a response-word algebra does not make field/action reconstruction statable.

I left #701 open: two of seven rows statable is the result, but the issue asks for the full matrix to be resolved, and five rows remain typed exits.

One thing worth flagging for your next brick, not a defect here. Your "known limitation" note is right that the counts range over the list literal rather than over #701's seven targets, and the anchors only pin rows 2, 3 and 5. Rows 1, 6 and 7 are prose-only — nothing breaks if pinned Mathlib later grows Lorentzian geometry or a Tomita conjugation, and the exit silently becomes stale. A cheap guard would be a kernel-checked receipt per absence-based exit, even something as blunt as a test asserting the cited Mathlib path still lacks the named declaration. Worth considering when a row's blocker starts moving.

PR #722, opened 2026-08-08
Add a conditional A2/A5 mechanism-variant example

jay-oraclizer:
## Summary

- Add a domain-neutral `MechanismVariants` kernel for variant-indexed traces, behavior cuts, authority views, support-dependent relations, exact-range encodings, and pairwise-versus-family comparisons.
- Add a finite conditional A2/A5 model based on Will's shared mechanism-design fixture and the OPH observable-normal-form and repair calculus.
- Export the generic kernel from the library root and include the complete example theorem surface in `AxiomAudit.lean` and both proof indexes.

## Authorship and provenance

Formalization and proof implementation by Jay Kim (`jay-oraclizer`), based on Will's shared A2/A5 mechanism-design fixture and the OPH calculus.

## Model and claim boundary

The fixture-to-Lean mappings are explicit, replaceable assumptions. In particular, the example uses two separate typed policy adapters:

1. a broad target with outcome observation; and
2. a flash target with hybrid observation.

Each adapter carries its own target, validity/core predicates, protected set, authority quotient, selected support, cost provenance, and comparison family. Under each declared policy, the finite model proves the pairwise order A2 < A5. The policies are not collapsed into a single interpretation.

The model also keeps observational and strategic statements separate. Protected outcome behavior can be preserved while a synthetic delay cost reverses the preferred strategy, so no observation-to-strategy implication is claimed.

The pairwise order is not promoted to a family minimum: an explicit third-variant counterexample has a smaller cut than A2. Accordingly, the contribution makes no family-wide, global, or unique-minimum claim. It also does not claim broad A5 safety or zero raw A2 capability growth; the authority result uses one declared quotient, and the raw-capability inclusion deliberately fails.

The repair results are fixed-relation statements. Full support is tied to the exact range of admissible traces and protected outcome representatives. A distinct missing-support relation supplies a static collar-hole no-go. This does not establish same-initial executable repair, all-path completeness, full `Repair` width, or a complete support family.

## Verification

- `lake build ObservableNormalForms` — success (8,262 jobs)
- `lake build ObservableNormalForms.Examples.AmdA2A5Conditional` — success (8,251 jobs)
- `lake build ObservableNormalForms.AxiomAudit` — success (8,260 jobs)
- Axiom inventory: only `propext`, `Classical.choice`, and `Quot.sound`; no unexpected named axioms and no runtime `sorryAx`
- Source scan: no `sorry`, `admit`, or `native_decide`; `git diff --check` passes

## Review focus

1. Whether the broad/outcome and flash/hybrid policy adapters capture the intended fixture interpretations.
2. Whether the generic kernel and finite example are placed at the right module boundaries.
3. Whether the library-root → `AxiomAudit` → conditional-example import path is appropriate.
4. Whether the attribution and conditional/nonclaim wording are precise.

Maintainer edits are welcome.


muellerberndt:
Thanks, I'll review ASAP :)

muellerberndt:
Thank again for the contribution!

Our prose paper *Observation-Determined Normal Forms* claims its machinery is substrate-neutral, and the shipped evidence for that claim is currently a single cellular automaton (`Examples/Rule90`). One example in one domain is a weak basis for a neutrality claim.

A Harberger auction contract is a genuinely independent second instance so this is a valuable contribution. 

Quick clarification on the A2/A5 nomenclature: This refers to @WGlynn's shared A2/A5 mechanism-design fixture. That reconstructed them from the transition systems:

- **A2: grace-period variant.** `initiate` moves `active incumbent` to a `pending` state, then either `complete` (buyer acquires) or `block` (incumbent retains). Both caller kinds permitted. The incumbent gets a window to respond.
- **A5: EOA-only variant.** A single `acquireEoa` step; no contract-originated acquisition trace exists at all.
- **baseline.** immediate acquisition by either caller kind.

Both are mitigations for the same threat: `BadFlash` marks exactly `baseline.acquireContract` bad, i.e. an atomic contract-originated acquisition / a flash-loan-style attack. A2 defends with delay; A5 defends by excluding contract callers.

## Maintainer edits

**1. Name collision with the alternating group A5.** In this repository `A5` overwhelmingly means the **alternating group A₅** (icosahedral symmetry): 16 modules under `Lean/Screen/` are named `A5*.lean`, and 34 files reference it. `A2` is also taken, by `Screen/A2HolonomyBridge.lean`. A file named `AmdA2A5Conditional.lean` in the same tree will read to future contributors as icosahedral-group work. This makes things confusing, so we'll add a short docstring paragraph naming the two variants and disclaiming the group reading fixes it.

**2. `FamilyVariant` reuses `.a2` and `.a5`.** The family counterexample operates on a synthetic type with a hand-written `familyCut`, entirely disconnected from the AMD model but its constructors share names with the real `Variant.a2`/`.a5`, so `a2_pairwise_lt_a5_in_declared_family` reads as a statement about the mechanisms. Your `PROOF_INDEX.md` row is precise about this ("a synthetic third variant"); only the code lacks the marker. A docstring on the type is enough.

The ~60 new declarations push the Lean count past 4200, so `paper-preview`, but that's a mini-fix.

PR #723, opened 2026-08-08
perf(A5SixAxes): band-split `rowF_mul` so the file compiles under a memory ceiling

velvetmonkey:
> “Yes, that would be nice, thanks!” — Bernhard on #718, replying to the offer of a band-split follow-up.

This is that follow-up. The diff changes one file, `Lean/Screen/A5SixAxes.lean`.

## What changes

The current `mul_closed` proof asks the kernel to decide closure directly through `Equiv.Perm` multiplication and list membership. This patch moves the exhaustive computation to raw value tables (`rowF`/`rowI`) with precomputed index tables (`mulT`/`invT`), proves bridges back to the listed permutations, and uses those bridges for `mul_closed` and `inv_closed`.

The quadratic `rowF_mul` check is split into twelve private five-row band lemmas (`rowF_mul_band0` … `rowF_mul_band11`, covering rows `0–4` through `55–59`), each with its own `decide`, and recombined by an exhaustive `rcases` ladder on `i.val`. A `set_option maxRecDepth 16384` is added where the deeper proof terms need it.

## Fresh measurement on the rebased branch

Base: `e91e3f4d` (`upstream/main`). Head: `6cd30111`.

Command, from `Lean/`: `/home/monkey/bin/leanbuild build A5SixAxes`, measured with `/usr/bin/time -v`.

- `rc=0`; `Build completed successfully (8256 jobs)`.
- `Built A5SixAxes (241s)` for the module itself.
- The full invocation took `2:02:22` because this worktree began without the pinned Mathlib packages or compiled oleans and bootstrapped the complete dependency graph before building the target.
- Full-invocation high-water RSS: `8,953,920 kB` (about `8.54 GiB`); zero swaps.

## What is unchanged and audited

- The proposition proved by the existing public theorem `mul_closed` is character-for-character identical to upstream; a statement-only byte comparison returns no diff. (`rowF_mul` itself is new proof infrastructure, not an upstream theorem.)
- All eleven `#print axioms` checks at the end of the file report exactly `[propext, Classical.choice, Quot.sound]`.
- A source search finds no proof use of `sorry` or `native_decide`, and no added `axiom` declaration.
- The twelve band declarations are private. The patch also adds public raw-table and bridge helpers, so it does not claim that the module's public name set is identical.

## How to verify

- `git diff --name-only upstream/main...HEAD` prints only `Lean/Screen/A5SixAxes.lean`.
- Compare the `mul_closed` proposition before the `:=` against `upstream/main`.
- Build `A5SixAxes` and inspect the eleven `#print axioms` lines at the end.


muellerberndt:
Reviewed and merged.

For a change advertised as performance-only, the load-bearing question is whether any proposition moved. I checked that mechanically rather than by eye: extracting every `theorem`/`lemma` statement from both revisions and diffing the sets shows **all 21 pre-existing statements survive byte-identically**, with nothing removed or weakened. `L60` itself is untouched, so `mul_closed` means exactly what it meant before. The 19 additions are 12 private band lemmas plus seven public table/bridge helpers.

Reproduced your measurement closely. `lake build OPHScreen` on a 24 GiB machine: rc=0, `Built A5SixAxes (120s)`, peak RSS **8.51 GiB** against your reported 8.54 GiB. Full umbrella: 8496 jobs, exit 0. All eleven `#print axioms` lines report exactly `[propext, Classical.choice, Quot.sound]`; no `sorry`, no `native_decide`, no added axiom. The theorem-count and `native_decide` inventory gates both pass — the count moves 4165 to 4172, since the twelve band lemmas are private and correctly excluded.

The band decomposition is sound by construction rather than by inspection: `0–4, 5–9, …, 50–54, 55+` partitions `Fin 60` exactly, and since the recombination is an exhaustive `rcases` ladder the kernel would reject any gap. That is the right way to build it — a coverage error cannot survive compilation.

The result matters beyond tidiness. The old single `decide` needed roughly 12–13.5 GiB for one declaration, which meant the file simply did not build under any cgroup ceiling below that, and a contributor on a 16 GiB machine would have watched it thrash indefinitely with no useful error. Dropping to 8.5 GiB and 120s makes the module reachable for people who could not previously compile the repository at all.

One provenance note, purely for the record and not a criticism: the band-split was originally offered by @qizwiz at the end of #718, and the "Yes, that would be nice, thanks!" you quote was Bernhard replying to that offer. You have been transparent about the lineage by quoting the exchange up front, and the implementation here is your own. Flagging it only so the credit is visible in the history.

PR #747, opened 2026-08-14
feat(QFT): stale-absence guards for the E4 status matrix

velvetmonkey:
Every typed non-evaluable exit of QFT/InheritanceMatrix.lean cites an absence that is true today and goes false silently when pinned Mathlib or the tree grows the missing structure; row 4 already went stale once on merge day and was corrected by hand in 4012dea5. Two guard layers make that staleness loud. Lean side (QFT/InheritanceMatrixGuards.lean): an elaboration-time scan over the full pinned Mathlib environment plus the matrix import closure that fails the build on any hit against the per-row absence probes of rows 1, 2, 3, 6, 7, asserts the row-1/row-2 presence citations by name, and fails closed on scan vacuity via module- and constant-count floors; the probe tables are linked to e4StatusMatrix by kernel-checked rfl receipts. Python side (code/audit/test_e4_absence_guards.py): whole-tree file scans the import closure cannot see (an unstarted QFT/TimeSlice.lean appearing anywhere, tree declarations for the cited absences), the comment-level row-1 Tomita TODO citation, the row-4 artifact presence citations, and a two-way probe-table sync check; unreadable or missing citation paths fail, never pass. The guards check that named probes are absent from the pinned surfaces; they do not prove any target unstatable by any route, and both files say so.


muellerberndt:
Merged as 895b5b55 after an independent three-part review that treated the guards as claims to refute. The soundness reviewer rebuilt the guard against the pinned Mathlib oleans (7872 modules, 529,757 constants scanned) and ran a nine-way mutation battery: a canary tomita declaration compiled into the import closure turned the build red with the exact row-1 finding, floor tampering and citation renames both turned it red, a deliberately colliding probe produced 287 findings, and the Python layer's TimeSlice and unimported-declaration canaries both fired. The mechanism is what the description says: a genuine environment enumeration with a hard error path and no silent-pass route, plus honest non-coverage statements in both layers. Good, careful work; the two-layer design with mirrored probe tables and a sync gate is the right architecture for this failure mode.

Maintainer fixes applied in the merge commit:

- The review found one real double blind spot, verified empirically: a probe-matching declaration placed inside the guard module itself passed both layers, because the Lean scan never sees the module being elaborated and the tree scan excluded the guard file. The tree scan now excludes nothing; declaration-name matching keeps the two meta-files from self-hitting, and the layer that watches the watchers watches itself.
- The absence-probe receipt checked a hardcoded row list beside the matrix; it is now computed from the live matrix shapes, so a shape change on a probed row breaks the receipt instead of drifting past it. The prose-only list stays declared, with its docstring saying so.
- A probe-charset receipt pins every probe to nonempty lowercase alphabetic, closing the dead-probe and sync-regex gaps the review identified.
- Probe vocabulary: rows 6 and 7 gain pseudoriemannian, semiriemannian, and waveoperator, each verified absent from the pin before landing. The suggested smatrix probe is dropped: it collides with fourteen basisMatrix-family declarations in pinned Mathlib and would have broken the build on arrival.
- A pinned-revision assertion compares the scanned checkout against the lake-manifest rev, so a local run against a stale Mathlib cannot report the absences as holding for the pin.
- lean-ci triggers now cover the row-4 artifact directories, closing the delayed-detection window for the one row that went stale before.
- Style pass per docs/STYLE_GUIDE.md: em-dashes out of both files, state-only wording (two test renames included), the intro claim narrowed to what the named-probe mechanism delivers, and one private-conversation reference replaced with the matrix's own terms.

Full Lean build (8549 jobs), the guard elaboration with the hardened probes, the extended test file (8 passed), and the standard mandatory suite are green on the merge commit; the invariant-mining lock was verified valid on the merged tree.

Issue #715, opened 2026-08-06
Proposal: algebraic extension layer for Infinite Observer (inductive-limit algebra) and Free-Will Agents (admissible instruments)


duswami:
This is a modest algebraic proposal layered on top of the existing OPH three-axiom surface. It introduces **no new axioms**.

### Two definitions

1. **Infinite Observer** — identified with the inductive limit  
   `A_∞ := lim→_r A_r`  
   of the directed system of accessible algebras already present in the published language (“global state on the inductive-limit algebra”), together with the capacity filtration `C_{r,i} = S(ρ_{r,i}) < ∞`.

2. **Free-Will Agents** — identified with the constrained subset  
   `W_{r,i} ⊆ U_{r,i}`  
   of already-declared update/repair instruments that leave the A1+A2-compatible set `K_r` invariant and preserve reachability of normal forms under the existing consensus operator `T`.

### What is claimed

- Every already-proved finite-observer theorem lifts under restriction from `A_∞` (or from compatible local families).
- The existing repair operators map `W` into itself under the same hypotheses used for confluence/termination.

### What is **not** claimed

The layer does **not** close any of the open quantitative physical-bridge or continuum-attachment obligations (Thomson transport term, physical `N` packet, family attachment, common-domain gravity tower, etc.). It only clarifies global-to-local structure and supplies an internal language for consistency-preserving local updates.

### Artifacts

Full definitions, white paper, workflow, Lean-style sketch, three tested finite-dimensional Python witnesses (consistent state, one-step repair, multi-step soft repair with measured contraction factor 0.6), one-command verifier (`verify_all.py`), and numerical receipts are collected here:

**https://github.com/duswami/OPH-Infinite-Observer-With-Free-Will-Agents**

Key files for quick review:
- White paper: `WHITEPAPER.md` / `WHITEPAPER.html`
- Testing guide + expected numbers: `TESTING.md`
- One-command check: `code/verify_all.py`
- Lean sketch: `lean/GlobalAlgebra.lean`

The contribution is offered for review, criticism, or possible absorption of the algebraic definitions into the main library. Feedback on whether the lift and preservation lemmas are useful (or already covered) would be welcome.

duswami:
Hello, im an amateur researcher, would be interested in anyone's perspective on this prospective addition the the OPH framework. :)

muellerberndt:
Thank, I'll check it out as soon as I can!

muellerberndt:
[...]

PR #749, opened 2026-08-18
E3 filtered-colimit layer: colimit observable algebra, limit isotony/…

velvetmonkey:
…locality, refinement independence, Cauchy class with time-slice, public-record subfunctor

Implements the open deliverables named in the LocallyCovariantLimit claim boundary (issue #700, per #715): the filtered-colimit observable algebra of a consensus tower over its directed regulator preorder, built as germs of private observables from the algebra_refine_refl / algebra_refine_trans receipts with NO injectivity assumed of any bonding map — the quotient is exactly by the refinement-null kernel (colimitMk_eq_zero_iff), and the concrete universal property is proved (colimitDesc, colimitHom_ext, factorization triangle).

Limit level: limitLocalAlgebra with isotony (limitLocalAlgebra_mono), locality (limitLocalAlgebra_commute), refinement independence (limitLocalAlgebra_refine), and the cocone receipt over the observable functor.  The Cauchy embedding class IsCauchyEmbedding is defined through the net's contravariant restriction maps (restrict inverts the isotony inclusion on both sides) and yields the time-slice property at the limit; nonvacuity (topToFull) and nontriviality (bottomToTop is not Cauchy) are proved on the support-graded example.  The public-record subfunctor publicObservableFunctor with its natural inclusion into observableFunctor and the commutative limit public algebra close deliverable 5.

The collapsing regime is exercised, not assumed away: collapsingTower is a committed ConsensusTower whose bonding map is not injective, its unit is a nonzero germ that dies in the colimit, and algebraRefine_injective_of_observer proves collapse is confined to observer-free regulators (state trace forces positive dimension; matrix algebras over C are simple).  The inclusion regime of the outside proposal is recovered as a theorem
(colimitMk_injective_of_refine_injective), not an assumption.

The new Cauchy embedding class intentionally ends the row-6 tree absence cited by the E4 inheritance matrix; row 6 is re-reviewed as its guard demands (order-theoretic Cauchy class now cited as present, remaining absences re-probed), with the Python companion table kept in sync.  No C*-norm/limit-norm completion and no limit-level restriction map are constructed; the claim boundary says so.

Build: lake build OPHConstruction (via leanbuild) exit 0, 8389 jobs; code/audit/test_e4_absence_guards.py 9 passed.  No sorry, no conclusion-True statements; axiom audit prints only propext, Classical.choice, Quot.sound.

### DIFF
--- a/Lean/QFT/InheritanceMatrix.lean
+++ b/Lean/QFT/InheritanceMatrix.lean
@@ -42,15 +42,16 @@ orphaning the matrix.
 | 3 | DHR / sector reconstruction | RESTRICTED SURROGATE COLLAPSES + exit (full DHR) | `QFT/SectorInheritance.lean`: every bijective localized endomorphism of the finite constructed net is inner (`privateSector_localized_inner`) and hence equivalent to the identity under the file's declared finite relation; nonvacuity by `sectorWitness_inner`.  This does not construct a DHR category or perform Doplicher–Roberts reconstruction.  Full DHR (quasi-local algebra, vacuum representation, transportability) is that module's typed exit. |
 | 4 | Field / action reconstruction | GATED BY DESIGN | Exact finite precursors now include the response-word algebra, the dense global Hodge comparison, and the genuinely local face-curvature/static scalar-vector action of `Screen/LocalFaceMaxwellAction.lean`.  The local action has the exact gauge kernel and conserved-source equivalence, but supplies no temporal dynamics, charge-current continuity map, source-selected physical current, physical matter/Spin action, spacetime carrier, or continuum limit.  Thus the full field/action reconstruction packet E4 would consume is still not supplied. |
 | 5 | KMS / thermality | PROVABLE (finite core) + REFUTABLE (net and public) + exit (horizon) | `QFT/StructuralInheritance.lean`: the finite algebraic Gibbs-KMS theorem on the witness private block (`privateGibbsState_kms`); the degeneracy countermodel — on the support-graded net every functional is KMS for every grading-preserving flow at every complex time (`supportGradedNet_regionalExpectation_kms_degenerate`); the cited B3 rigidity refuting public thermality (`publicRecord_no_thermal_flow`).  Horizon thermality is that module's typed exit. |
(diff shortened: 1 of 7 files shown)


velvetmonkey:
---

## CI status: one red is ours to report, not ours to fix

The Lean side is green: `build` passes, and `mandatory collection` passes on
both `ubuntu-latest` and `windows-latest`.

`paper-preview` fails, and the cause is this branch working as intended:

FAIL README.md:115: floor 5800 is stale (expected 5900, count 5909)
FAIL README_FR.md:135: floor 5800 is stale (expected 5900, count 5909)
FAIL extra/compact_proof_of_oph.tex:530: floor 5800 is stale (expected 5900, count 5909)
Lean public theorem/lemma declarations: 5909 (expected floor: 5900)

The E3 layer takes the public theorem/lemma count past the stated floor of 5800.

**We attempted the bump and then reverted it, deliberately.** Commit `47ca47a2`
applied `tools/check_lean_theorem_count.py --fix` output unedited, one number
per file. Two gates refused it, and both refusals look correct to us:

1. `claim-registry / validate` treats `README.md`, `README_FR.md` and
   `extra/*.tex` as public claim surface, and requires a matching `claims/` or
   `tools/check_claim_registry.py` change. That gate exists so claim language
   reaches human registry review, and an outside contributor silently restating
   how many theorems the library contains is exactly what it should stop.
2. `paper-preview` additionally requires the rebuilt
   `extra/compact_proof_of_oph.pdf` and `paper/paper_release_manifest.json` to
   be committed.

So `add1dd84` reverts it. **The branch tree is now byte-identical to the
reviewed brick `5d98e8a6`** (`git diff --stat 5d98e8a6 HEAD` is empty). The
floor bump, its registry entry and the artifact rebuild are a maintainer act,
and we would rather leave one legible red than edit a claim about your paper on
our own authority. Happy to do it in a follow-up if you would prefer.

### Separate finding, offered as a courtesy

`tools/check_lean_theorem_count.py --fix` writes a commit that cannot pass this
repository's own CI. It rewrites the three stated floors and performs neither
downstream obligation: no claims-registry touch, no artifact rebuild. Anyone
who crosses a floor hits the same wall, and the `--fix` flag reads as though it
finishes the job.

muellerberndt:
Thanks! Don't worry about the paper build issues, I'll take care of those.

PR #746, opened 2026-08-14
feat(lean): formalize protected-behavior obstruction profiles

jay-oraclizer:
## Problem

A protected record can have a nonempty consistent fiber while still failing construction under the declared scheduler. This change formalizes the first-hit obstruction profile needed to separate four distinct questions: target existence, positive reach from some declared source, almost-sure reach from every declared source, and selection of one silent-equivalence endpoint class.

## Result

- Defines the canonical path-weight first-hit law and four nested layers with complementary obstruction cuts.
- Proves T0-T8, including structural derivation of scheduler support/rewrite correspondence at T6.
- Makes T7 the direct formal dependency for the protected-behavior cut corresponding to #722.
- Proves exact qualitative and quantitative finite-state transport under the declared morphism packet.
- Activates native TwoBit, C4, C6, and nonidentity fixtures plus explicit countermodels.
- Adds a deterministic finite explorer with baseline and deliberate-mutation obligations.

Finite Markov-chain reachability and closed-class facts are standard ingredients. The contribution here is their source-quantified integration with protected observations, exact scheduler/rewrite hypotheses, behavior-cut recovery, morphism transport, and native OPH fixtures.

## Publication surface

The observable-normal-forms component paper contains the detailed definitions, theorem, proof sketch, examples, and nonclaims. The flagship paper contains only the minimal theorem connection, corollary, and scope boundary. Maintainers may revise or exclude either manuscript commit independently of the Lean/claims commits.

The Isabelle companion is a separate `PARTIAL / NO SAME` PR. It provides assumption-transparent HOL consequences and bounded controls, but does not re-prove the kernel-derived first-hit law, scheduler correspondence, or quantitative pushforward.

Author, affiliation, correspondence, and contribution metadata are unchanged.

## Explicit nonclaims

This change does not establish a hitting-time rate, expected hitting time, mixing theorem, infinite-tower result, deployed implementation refinement, product safety result, or model-to-code equivalence. The four cuts form a product preorder, not a scalar score or total order.

## Commit topology

1. `1af03757533f5abd9f486e961d173d96ab6b0f7e` - Lean theorem, examples, imports, AxiomAudit, standalone package, explorer.
2. `6935957bc7e36323384bfd5e7a5472d4dbd792b4` - component/flagship integration, PDFs, bibliography, release manifest, minimal discovery surface.
3. `9aa3b201f5d3e473c5511dac3102af07ecc43811` - current claims, assumptions, dependencies, falsification/novelty rows, guards, workflow, tests.
[...]

PR #752, opened 2026-08-21
Add source-derived phase-effect selection

jay-oraclizer:
## Summary

- derive the finite phase-effect family from the permitted source fields
- use a target-free deterministic selector
- prove the selected effect equals `sourcePhaseLift`
- connect the 48/12 source relation to the 36/12 canonical bridge
- add deletion-sensitive and source-relation mutations

## Context

This continues the line of work in #722 and #746. Those contributions isolated the protected-behavior and finite first-hit/transport layers at the abstract level. This PR addresses the next source-side question in the EventAlgebra lane: whether the phase effect consumed by the current Lüders instrument can be generated from the repository's permitted source fields rather than supplied as a declared premise.

The answer is positive at the effect-value level. A target-free deterministic selector produces the exact declared phase effect. Event witnesses are not unique: eight source events produce the same +Y effect, so the result deliberately claims effect equality rather than witness uniqueness.

This is research lineage context, not a theorem dependency claim. The new modules do not import or depend on the theorems added by #722 or #746.

## Boundary

- does not construct a new Lüders instrument
- does not source-produce an instrument
- does not discharge PR-03 or full PR-64/PR-65
- does not provide preparation, public outcome, readback, provenance, custody, measurement, or physical validation
- event-witness uniqueness is false; uniqueness is only after effect-value deduplication
- does not modify any manuscript or PDF

## Verification

- `python code/born_context_phase_lift/source_phase_selection_family.py --payload code/born_context_phase_lift/BORN_CONTEXT_WEB_PAYLOAD.v1.json --output <temp>/source-phase-selection-packet.json --verify-mutations`
  - PASS: 60 cells, 48 enabled, 12 disabled, 36 state-changing, 12 equality cases
- both generated Lean modules reproduce byte-for-byte from the source packet
- `python tools/test_source_phase_selection.py --payload code/born_context_phase_lift/BORN_CONTEXT_WEB_PAYLOAD.v1.json --module-dir code/born_context_phase_lift`
  - PASS: selector `pair:00-02`, two effect values, event multiplicity 8/4, 48/12, 36/12, all six mutations
- `lake build EventAlgebra`
  - PASS: 8,315 jobs, including the current Lüders instrument and source-selection theorem
- claim-registry projection independently recomputed from `8dac48260677d12b39870ee12b6d7b73d2ce9ad9c707e37a8fdd351c0cf4b295` to `32a037fe56716be06d99d63ea6b775c7ff9ec87133e0d1a6c48037878cebf8ac`
  - checker pin matches; an explicit premise-edge mutation is rejected
- claim registry, premise register, observation ledger, generated surfaces, dependency graph, and exact 19-file inventory: PASS
- `python tools/run_mandatory_suite.py`
  - PASS in standard mode; the repository-defined nightly heavy steps remain deferred to `--full`
- exact staged public-surface scan: credentials, private paths, AI attribution, proof debt, internal process identifiers, caches, and TeX/PDF changes all zero

## Review target

- technical theorem and target-free generator
- PR-04 axiom-removal proposal
- effect-selection conjunct of PR-65
- related to issue #730


muellerberndt (approved):
Reviewed the Lean modules, the generator, the registry row, and the premise-register notes; reproduced the packet build with mutations and the test tool locally; reader-style and claim-registry gates pass on the branch. The theorems are correct and the boundary is stated carefully. One scope point is added in maintainer follow-up commits: the orientation of the selected effect (+Y rather than -Y) is fixed by the commutator order P_right P_left - P_left P_right and the lexicographic pair index, both producer conventions; what the source fixes convention-free is the unordered pair of Y projectors. Merging.

PR #757, opened 2026-08-27
Fold OphGap Lean proof into the tree: Python produces data, Lean justifies the argument

velvetmonkey:
This branch folds the complete OphGap Lean proof, its kernel-checked transport from the BFS-relabelled graph to OPH’s original carrier ids, the T1/T2/T3 slices, build wiring, and data generators into the OPH Lean tree, while removing six machine-local path references from the contribution.

The gap is proved for ONE frozen graph, identified by `domain_freeze_sha256 = a0be6fc64aecf9ca375fd91c57315e8af5e5cf161c99611f4844ba8f452ae7ff`. It is not a general theorem about all graphs, all OPH data, or all source-gap claims.

Python now only PRODUCES data files; no step of the ARGUMENT is justified outside Lean. That is the actual advance over the pointer merged this morning: the relabelling/transport conditions and the resulting statement about OPH’s original carrier ids are checked by Lean’s kernel rather than being justified by the Python step.

What is still NOT verified is the extraction that produced the frozen literals. Those literals remain the proof’s input.


velvetmonkey:
The claim-registry gate is path-based, so it fires on any README edit.

Our entire claim-surface diff is two lines, both the machine-generated Lean
theorem count:

  - More than 9700 Lean theorems with
  + More than 9900 Lean theorems with
  - Plus de 9700 théorèmes Lean sans
  + Plus de 9900 théorèmes Lean sans

claims/public_surface_quantitative_claims.json already lists that number under
nonclaim_numeric_contexts, naming tools/check_lean_theorem_count.py as its
checker. So the gate at .github/workflows/claim-registry.yml:49 is stricter
than the policy it enforces.

Two options, both yours: exempt a change whose whole diff falls inside a
declared nonclaim context, or let check_lean_theorem_count.py --fix touch the
claim surface. Happy to do either if you say which you prefer.

We deliberately did not edit claims/ to clear the red, since the gate exists
to force a human look at claim changes and satisfying it silently would defeat
that.

PR #767, opened 2026-09-02
Identify the antipodal port action with the six-axis PSL(2,F5) model

gift-framework:
## Summary

This PR closes a finite representation-theoretic bridge between two existing OPH constructions.

`A5PortAction` lists the sixty proper incidence automorphisms of the twelve-port carrier and proves that every row commutes with the antipode. Quotienting the twelve ports into six antipodal pairs therefore produces a six-axis action.

Independently, `A5SixAxes` contains the repository's explicit sixty-row six-axis `PSL(2,F5)` model.

This PR proves that, after one fixed explicit relabeling of the six quotient axes, the complete antipodal port action is exactly the `A5SixAxes` action, with an explicit bijection of the sixty rows. It also proves that the quotient action is faithful.

This supplies a concrete bridge for the existing golden-sector character calculations, which are evaluated on the same port-action rows.

## Scope

This is a finite-action equivalence theorem.

It does **not** formalize an isomorphism to Mathlib's abstract `PSL(2, ZMod 5)`, prove the abstract classification `PSL(2,5) ≅ A5`, construct the binary icosahedral group `2I ≅ SL(2,5)`, establish the McKay correspondence, select the golden ratio as a physical parameter, derive any mass formula, or identify the carrier action with physical rotations.

In particular, this PR does not derive or promote `27^φ`.

## Checks

- antipodal quotient well-defined
- explicit six-axis relabeling and inverse
- explicit bijection of all sixty rows
- pointwise equality on all 60 × 6 row/axis pairs
- converse realization of every six-axis row
- faithfulness of the quotient action
- no proof debt
- `lake build A5PortSixAxesBridge`
- full `lake build OPHScreen` (8,465 jobs)
- structural regression gate (8 passed)
- active-surface inventory regeneration and check
- mandatory suite (standard profile; passed)


muellerberndt:
Merged as b8404d67. A note on why: the bridge is load-bearing rather than decorative. It is anchored through P60_matches_port_rows to the exact rows the golden-sector characters evaluate, so the six-axis reading of those calculations is a kernel-checked equivalence with an explicit relabeling and an explicit row bijection, and the faithfulness theorem rules out a hidden kernel in the quotient. The scope paragraph is exactly right, including the explicit non-promotion of 27^phi. Registry, inventory, and test surfaces were verified locally before merge.

PR #768, opened 2026-09-02
Add external Arithmon negative control for Koide balance

gift-framework:
## Summary

This PR adds a pinned external negative control for the exact OPH Koide balance.

OPH proves that the balanced nonnegative circulant has \(Q=2/3\), with the positive chamber required for the physical square-root-mass reading. Independently, public Arithmon receipts provide the version-pinned mass pair
`m_mu/m_e = 27^phi`, `m_tau/m_e = 3477`, and `Arithmon/K7-Lean` proves the strict result
`Q(27^phi, 3477) < 2/3`.

KOIDE-CROSS-0 pins those external sources and recomputes the same scale-invariant Koide quantity locally. No experimental lepton masses are consumed.

Two mutation controls replace either the tau or muon ratio by the corresponding exact-balance root and recover \(Q=2/3\), checking that the incompatibility detector is load-bearing.

## Boundary

This is a compare-only external-framework control, not a prospective prediction freeze.

The OPH and Arithmon results are kernel-checked in their respective repositories but are **not composed in one Lean kernel** here. No Arithmon statement is imported as an OPH axiom, no charged-family physical attachment is derived, and the result does not adjudicate which framework is physically correct.

The conclusion is only that the version-pinned Arithmon mass pair cannot simultaneously be an exact balanced positive-chamber OPH charged-family spectrum.

## Checks

* deterministic artifact regeneration / stale-output check
* exact external source pins
* zero experimental-mass input
* local sign agrees with the pinned Arithmon theorem
* balance-root mutation controls
* regression coverage through the existing Koide test surface

### DIFF
--- a/code/particles/data/arithmon_koide_cross_0.json
+++ b/code/particles/data/arithmon_koide_cross_0.json
@@ -1,0 +1,37 @@
+{
+  "schema": "oph.external_arithmon_koide_cross_fixture.v1",
+  "role": "external_framework_compare_only_version_pin",
+  "freeze": {
+    "arithmon_k7": {
+      "repository": "Arithmon/K7",
+      "commit": "af9224cf3bd0c7bb8a895b98d4d1146eb138b555",
+      "path": "docs/wiki/Observable-Reference.md",
+      "git_blob_sha": "646500c22e40fcd89d3584df7d64d5a9d452f754"
+    },
+    "arithmon_k7_lean": {
+      "repository": "Arithmon/K7-Lean",
+      "commit": "836296db191c49307f7df59aa0df989399a0b1bb",
+      "path": "GIFT/Relations/KoideAssembly.lean",
+      "git_blob_sha": "47ff66d563e50f64f73ad085490459f28ff0df85",
+      "theorem": "GIFT.Relations.KoideAssembly.koideQ_gift_lt_two_thirds",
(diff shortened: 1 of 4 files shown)


muellerberndt:
Merged as e4b7d8cd, with a note on why, since cross-framework material is not something this repository normally takes.

The value is not the comparison itself. The control stress-tests the OPH proof machinery at its weakest point: the pinned pair sits within 0.15 percent of the empirical ratios, so the sign of Q - 2/3 is decided at the fifth decimal, and the artifact earns that sign with margin. The recompute runs 120-digit decimal arithmetic in a local context, the balance-root mutation controls restore Q = 2/3 to 1e-110 and prove the incompatibility detector load-bearing, and the frozen artifact carries a stale-output check. During review the external pins were fetched and verified at the pinned hashes: both commits exist and the pinned blob contains koideQ_gift_lt_two_thirds with its interval proof, so the fixture is auditable rather than decorative.

The boundary is what made this acceptable: compare-only, no composition in one kernel, no Arithmon statement imported as an OPH premise, no physical adjudication, and no paper or claims surface touched. A negative control of exactly this shape adds falsification surface without adding premises. That is the standard any future external-framework material will be held to; this merge sets no precedent for comparative-framework claims in papers.

Issue #777, opened 2026-09-09
[V4-S2] The read law as a production: routing record-metric reads on the support wiring


muellerberndt:
## Audited full-family result — 2026-09-21

The complete conditional routing contract is satisfied by the reviewed #910/#913 contribution, merged with the separate #922 native arithmetic/publication result in #924 at [`752b1341`](https://github.com/FloatingPragma/observer-patch-holography/commit/752b134199e455fc1d8714a53147e06af277d8bf). This uses the **declared-M1 exit**, not a derivation of M1. The earlier September 13 open-work description below is historical and superseded by this result.

- All four q13/q21 baseline/source-intervention histories regenerate against retained decoded phase and whole-stream hashes. Independent semantic replay checks every operation, consumed writer, complete metric menu and recurrence: 19,113,548 events per q13 run and 292,722,053 per q21 run, 623,671,202 in total. The induced logical value-consumption order and integer intervention theorem have finite Lean proofs under the checked transport certificates.
- An analytic all-level graph argument composes separate Lean lemmas. The live-store theorem gives 23C scalar slots under explicit lifetimes; the executable occupied-host bound is 14C+9n. Four q3 mapped controls retain 91,272 events. The q13/q21 slot applications are bounds, not new mapped executions or measured RSS.
- The source-law premise remains complete and explicit in PR-52 and the papers. Its derivation/justification is transferred to #779 and #740: protected records and zeros, export/reset/factor-two capture, local memory and accumulation, commits, placement, full read menus, routing/controller and observer interface. Those issues retain their broader physical contracts. No physical premise or observation rung is promoted.
- #922 separately gives reusable mean-only logical accumulation and publication on singleton compatible-payload sets. It preserves values at attenuated gains; it does not supply immutable general archives, physical precision or a full native q13/q21 compiler.

Every intermediate operation is retained and charged. Semantic order does not remove serial/control ancestry. Regeneration validates the computational streams, not historical host execution, original compressed archive custody or a reproduced memory measurement. Auxiliary events inherit neither a physical-clock identification nor the distinguished logical count-volume weight. The closed-sum reset obstruction remains restricted to its stated grammar.

The flagship, spacetime paper, scientific registers and five affected book files carry these boundaries. The full source-selection and physical-world tasks remain open in their owning contracts.

---

Derive and compile the record-metric read law on local support wiring, including repeatable versions and complete resource accounting.

## Current audited frontier — 2026-09-13, second tranche

[...]

PR #917, opened 2026-09-20
Derive reusable record transport on captured W12 seams toward M1

MarioPoneder:
The encoded copy/reset primitive does not itself provide reusable transport between actual carriers. This change constructs a finite four-level record bus on captured W12 seams, including native approximate cleanup and explicit rounding and readout error budgets.

## Objective and deliverables

- A checked route from carrier 0 to neighbouring carrier 3, via two disjoint three-edge paths; the positive path crosses carrier 1. Every executed mean is a captured seam.
- Two prepared records with payloads in `{-3/2,-1/2,1/2,3/2}`, read in order `[0,1,0]` using the same 12 scalar registers. All 16 payload combinations retain complete tapes: 27,264 means, including all cleanup and consumed writers.
- Twenty-five Lean theorems establish native compilation, seven-mean cleanup with contraction `7/8`, repeated cleanup, retention and composed receiver error bounds, operation counts, readout separation and integer rounding bounds. All dependencies are transitively audited against the standard axiom allowlist.
- An independent verifier reconstructs captured edges, schedules, exact ideal linear maps, rounded integer execution, writer ancestry and resource counts derived from the validated execution. It enforces canonical artifact bytes and records arithmetic maxima. Each history uses 1704 means, including 729 cross-carrier means, on a `2^-20` grid. The minimum certified decoding margin exceeds `0.0150217` under the declared allowances.

## Exit and scope

The [contract](https://github.com/FloatingPragma/observer-patch-holography/blob/3a181999808a5839c61cc1c4606b82ab62c37eb9/code/source_reusable_bus/CONTRACT.md) is met within its finite declared interface; the [audit](https://github.com/FloatingPragma/observer-patch-holography/blob/3a181999808a5839c61cc1c4606b82ab62c37eb9/code/source_reusable_bus/AUDIT.md) checks each deliverable and assumption.

The 12 counted registers are an active subsystem of the 15,360-port support; spectator storage and physical isolation are outside the accounting. The source amplitudes attenuate. Cleanup leaves bounded ideal cross-record influence, and every raw ancestor is retained. Correct four-level decoding does not establish immutable scalar archives, arbitrary writes, newly created versions, unlimited reuse or absence of physical causal influence. Preparation, isolation, codebook, schedule, addresses, comparator and additional error limits are supplied. Rounded means are a bounded implementation model and can change total load by a grid unit; no physical rounding mechanism or measured noise limit is derived. Cleanup contracts balanced amplitudes, while common-mode errors remain in the error budget. Storage widths apply to the retained integer model, not arbitrary real perturbations or Python object memory.

This is a finite route on a declared capture. It does not close M1 or #777/#779/#740, compile q=13/q=21 histories or supply the physical clock/count/common-world joins. Paper content, existing scientific receipts and the frozen mandatory runner are unchanged.

This PR is stacked on #915 to use its encoded-memory theorem without duplicating that review. Proofs, executable controls/CI, and documentation/audit are separate commits, followed by an audit-corrections commit.

## Validation

- Full `lake build Geometry` passes (8427 jobs). The direct 25-theorem axiom audit passes; an isolated nonstandard-axiom probe is rejected.
- All 24 stages of mandatory shard 0 pass on Windows and Linux, including the complete Lean comment/axiom-coverage gate and historical-source checks. External custody verifies on Linux; Windows reports its existing unsupported parser.
- 215 tests pass per platform: 97 bus controls, 64 encoded-memory controls, and 54 CI/sharding regressions. Both artifacts reproduce byte for byte.
- The scientific Python suite collects 5348 tests successfully; this is a collection check.
- Six exact adversarial trajectories maximize signed error at each read and check idle-register drift and decoding. Actual CLI invocations reject altered proofs, tapes, receipts and noncanonical bytes; controls reject disabled CI audit targets. Other controls reject invented seams, missing cleanup, assigned resets, stale writers, changed read interfaces and understated resource/error budgets. A two-fractional-bit execution loses the first record; midpoint observations are ambiguous.
- Prior encoded-memory, feedback-transport and captured-routing JSON files remain byte-identical. `git diff --check` passes, and a virtual merge with #914's green head succeeds.


muellerberndt:
Merged on `main` through the stacked merge commit 04f8fac4 (integration commit fceefd9c, tip a955996d). GitHub cannot mark this pull request merged because its base is the contributor branch below it rather than `main`, and retargeting is refused once the head is fully contained in `main`, so it is closed here as merged-by-hand. The review findings live in the maintainer commit and the issue 777 comment.

PR #918, opened 2026-09-20
Derive native sum records and polynomial chain cleanup

MarioPoneder:
Prepared-record transport could not create a new record, and its cleanup theorem covered only a four-cell bus. This adds a seven-mean native write of C=A+B on captured W12 seams, followed by C,A,C,B reads that preserve the old logical values through explicit scale tracking. All 16 input pairs and all 36,464 scalar means are independently replayed.

The general finite-chain proof gives residual at most A*(n+1)^2/(2n+1)*2^-k after (1+2n)*(n+1)^2*k scalar means. It includes untouched archives, the zero-edge boundary, forward attenuation and operation counts. A raw-range obstruction and a supported schedule-dependence control retain their precise limited scopes.

Validation:

- 39 new Lean theorems pass a transitive standard-axiom audit; full Geometry build passes.
- 359 focused tests pass on Windows and Linux, including 144 native-update tests, resealed event mutations, an alternative valid program, real source-file tampering, CLI negative controls and perturbed executions.
- All 24 stages of the frozen standard mandatory shard 0/6 pass on Linux, including external custody/history checks. An indirect injected axiom is rejected.
- Every finite read has a positive certified error margin; the minimum exceeds 0.0032996 under the stated error hypotheses.
- Compact artifacts total 156,363 bytes and 3,637 lines: one complete inspectable trace plus independently expanded and checked commitments for every input history.

The maintainer-style follow-up audit corrected the stale published theorem-count floor (11,022 declarations), added the missing imported audit-command source pin with a recursive dependency-coverage check, and verified actual low-precision decoder decisions (12 wrong answers and 16 ambiguous reads at Q=2^8). The scientific traces, ideal formulas and positive precision bounds are unchanged.

Follow-up to #917; related to #777. **This does not fully derive M1 or close #777.** General record programs and captured routing embeddings, complete native q=13/q=21 histories, their intervention-preserving refinement, and selection of the metric-neighbour read/controller law remain open. The schedule control is not a full A1--A3 countermodel. No scientific obligation is transferred or administratively closed.

The contract, reproducibility commands, resource boundaries, audit corrections and remaining obligations are in `code/source_native_updates/{CONTRACT,README,AUDIT}.md`. Proofs, executable evidence and audit documentation are separate commits.

PR #919, opened 2026-09-20
Derive constrained read-selection criteria and the observer-cover boundary

MarioPoneder:
This establishes when finite classical information projection can guarantee a remote record read. It removes the uniform/IID/full-simplex restrictions from the earlier schedule argument and connects the general result to the actual scalar transition law. It also proves why an injective observer cover alone is insufficient to apply a complete-history support argument.

For any attained weighted classical KL minimum over a convex nonnegative feasible family, with faithful reference and positive weights, every feasible positive atom remains positive at the optimizer. A zero-failure event, or a nonnegative failure observable of the scored atoms, is therefore already excluded by the entire feasible family. The native-read bridge proves that a guaranteed complete-history read requires every feasible schedule to exclude every cut-avoiding history, even when the decoder sees the complete receiver transcript and the whole word.

The cover boundary is substantive: on one convex three-history family and one faithful compatible reference, two state-determining covers have different unique minimizers, with history-0 probabilities zero and 1/6. Lean proves both minima over the entire family. Reconstructing the missing history probability through the coarse cover requires a negative offset, so it does not meet the positive-readout criterion.

Executable evidence includes an exact nonuniform correlated optimizer, the same-reference cover comparison, nonconvex and assumed-success-face controls, all 2,186 small-chain histories through length six, and positive/negative transport controls on the captured W12 support. The full captured census gives 46,050 seams and an eleven-edge source cut. The uniform specialization has cut-avoidance probability above 58% after 2,279 attempts and balanced-input error above 29%; that numerical bound is not asserted for arbitrary constrained schedules.

Validation:

- 48 Lean theorem declarations pass the transitive standard-axiom audit; full Geometry builds. An indirectly injected extra axiom is rejected.
- 544 focused tests pass on each of Windows and Linux, including 185 selection tests, real source-file tampering, resealed semantic forgeries, feasible-but-nonoptimal optimizer certificates, invalid covers and actual CLI failures.
- An additional audit individually corrupted all 789 scalar evidence fields, resealing history hashes where applicable: every corruption was rejected and the unchanged valid evidence remained accepted.
- Registry, theorem-count, Lean comment/prose style, portability, provenance, Lean-CI-budget and whitespace gates pass. The frozen runner and existing scientific evidence are unchanged.
- The two evidence files total 16,025 bytes and 535 lines. Compact commitments are independently expanded and verified, not accepted as proof by hash.

The classical support property is established mathematics, independently formalized here and applied to M1; it is not claimed as a new entropy principle. **M1 remains unproved and #777 remains open.** The source must still generate the complete feasible grammar, faithful reference, actual observer cover and accepted-record output/stop policy. No full A1--A3 countermodel or noncommutative selection theorem is claimed. Complete native q=13/q=21 execution and operational refinement remain open. No missing premise is silently adopted or transferred.

Stacked on #918. Mathematical results, executable certificates/CI, and the scope/audit documentation are separate commits. The objective, deliverables, exit and detailed audit are in `code/source_read_selection/{CONTRACT,README,AUDIT}.md`.


muellerberndt:
Merged on `main` through the stacked merge commit 4b8c9e4e (integration commit fceefd9c, tip a955996d). GitHub cannot mark this pull request merged because its base is the contributor branch below it rather than `main`, and retargeting is refused once the head is fully contained in `main`, so it is closed here as merged-by-hand. The review findings live in the maintainer commit and the issue 777 comment.

PR #925, opened 2026-09-21
Derive finite bank execution and noisy publication from pair means

MarioPoneder:
Computed outputs can be stored, consumed by later layers and retired after their last admitted read using scalar pair means. The two-bank compiler implements `v[t+1,i] = 1 + sum(v[t,j])`, including duplicate requests, empty sums, negative values and bank reuse. Lean derives the general compiler's recurrence, native lowering and noisy local publication; completed-program correctness is not a caller-supplied certificate.

The formal result covers arbitrary finite typed layer programs. It derives input-bank preservation, workspace and retirement invariants, public magnitude caps, exact native word lengths and a whole-history error ledger. Explicit permutations provide a declared shared bus, and finite cleanup/grid choices discharge the strict decoder-margin premise. `SourceBankExecution.finite_execution` combines these results: for every finite program and public amplitude bound there exist payload-independent precision choices such that execution within the stated preparation, per-mean and sampling bounds publishes the layered recurrence.

The captured implementation has a separate independent verifier. It reconstructs every native edge block, repetition, protected version, orientation, exponent and cap, then executes rounded means and checks a local interval decoder without supplying the expected answer to it. A compact generated Lean certificate matches the captured signed control's complete 42-stage logical stream, every observed scale, per-stage work, total of 1,798,154 means and all nine final output pairs. Kernel-checked negative cases reject empty/truncated/extended traces, wrong instructions, undercharged work and incorrect scales. The exporter freshness gate rejects stale correspondence and forged inputs through its real CLI.

Evidence is compact and reproducible:

- Nine signed three-layer histories exercise computed-value feedback and bank reuse.
- Two complete 27-site golden-menu histories execute both layers, including a centre-site intervention. Their words total 1,409,653,226 means: 292,557,295 are explicitly evaluated and 1,117,095,931 stationary effects are certified after a complete-block fixed-point check. Every repetition remains charged.
- All 45,940 two-bank routes at populations 27, 2,197 and 9,261 are independently checked, together with every metric-menu decision. The large families have explicit sufficient work and precision bounds. Raw route and execution witnesses are regenerated outside git.

[...]

Issue #740, opened 2026-08-12
[V3-CW1] Common-world integration: one inhabited architecture


muellerberndt:
## V4 integration checkpoint and retained physical obligations

#730 owns the first common source-family scalar/Hilbert/detector packet. Consume that packet incrementally while #777/#778 address source selection; keep #740's full terminal contract below. The packet needs one compatible radius/refinement, action, temporal inertia and clock, not a product of unrelated witnesses. #754 owns physical comparison and empirical units. #776 is a wiring diagnostic, #782 formalizes the limits, and neither by itself supplies source-selected matter dynamics or operational boosts of histories.

Closing #777/#778 with a supplied remainder must not be read as producing their physical inputs: preserve each retained premise here and in every direct scientific consumer. Final integration must also attach #780's current to the same matter action and sourced family, and #781's recovered specification to that realization.

Public derivation maps must agree with these scientific scopes. In particular, a Hilbert/GNS construction is not yet a common dynamics/region map, and finite order plus a dimension statistic is not a manifold-existence theorem. The September 10 review found those stronger readings in the linked learning chapters; align them when publishing the corresponding research claims.

Integrate the physics lanes on one concrete, explicitly identified scientific
configuration. Separate conditional witnesses on unrelated carriers do not
establish common-world adequacy.

## Operational cone selection and count clock

The spatial pieces compose on the same source carrier: the complete-census
repair response selects the rank-three Gram metric, its unit directions give
the source sphere, and a scalar axis gives the compatible Lorentz module.
Under natural feasibility/objective and unique-minimizer premises the uniform
directed-seam spatial operator is already selected.

- [x] **Conditional cone rigidity:** a nonzero closed convex pointed cone,
  invariant under the finite irreducible carrier rotations and one continuous
  operational boost direction, is exactly a future or past Lorentz cone.
  Time orientation chooses future. The full finite-matrix product proof is
  analytic; an algebraic frame action does not establish operational covariance.
- [x] **Retrospective count clock:** on the declared golden-source order/count
  family, (|I|/|J|)^(1/4) tends to the proper-time ratio for fixed interior
  timelike diamonds. The decoder uses authenticated ancestry, with no spatial
  coordinates, timestamps, density or oscillator. Positive volume-only
  inertial additivity selects the fourth root up to units. Finite error
  enclosures and resolved timelike-curve accumulation have analytic proofs.
- [x] **Exact decoder evidence:** a fresh 500-event q=5 execution gives counts
[...]

PR #947, opened 2026-09-21
Derive connected native recovery and temporal preservation constraints

MarioPoneder:
The stored-program construction executes supplied read requests. This change derives which meanings native observations can identify, constructs complete recovery from connected topology, and derives a temporal admission rule that preserves the ability to finish. It builds on the merged source-selection and stored-bank results in #919, #925 and #927.

### Objective and result

For a finite connected scalar support, one receiver can recover every initial scalar using exactly N calibration samples and at most N(N-1)/2 supported pair means. The successful route is constructed from connectivity; it is not supplied as a premise.

For a specified linear meaning, the past observations and current state give a necessary and sufficient continuation test through their joint kernel. A guard admits a proposed mean exactly when the record/state map remains injective on the specified source-record domain. One topology-derived completing word is accepted from every protected checkpoint whose retained record determines its current receiver scalar and completes recovery whenever it occurs among arbitrary other proposals. Under an explicitly supplied uniform full proposal law, the kernel-checked failure count is at most `(m^B-1)^k` against the original `m^(Bk)` denominator. Rejections remain charged.

### Deliverables

- 98 kernel-checked theorems covering native observation factorization, connected completion, preservation, proposal counts, full-state erasure, and the composition with constrained full-history KL selection. Arbitrary nonlinear meanings on finite independent product domains also have a proved unique minimal direct-read set.
- Independent primal/dual certificate verification, receiver-only native decoding, and exhaustive raw/guarded censuses of all 9,841 three-edge-chain words through horizon eight. The verifier does not import the producer.
- A complete, independently checked topology plan on the captured 15,360-port support: 425,979 means, supported simple paths, calibrated relays, complete coverage and explicit work. This large plan is topology evidence; eight five-port histories separately replay payload recovery. Retained controls and receipt total about 31 KB; expanded tapes are regenerated.
- A standalone paper derivation, rebuilt PDF/manifest, claim registry entry and Windows/Linux verification workflow. The commits separate proofs, executable evidence, publication integration and audit corrections.

### Validation and scientific exit

CI passes on `b312639cef5d455afa1857bd9000f759172aad8f`: all 46 applicable checks succeeded, with no pending checks; the scheduled-only nightly job is skipped as configured. The branch is mergeable against current main.

- The transitive axiom gate passes for all 98 added theorems using only the standard Lean axioms.
- All 71 temporal tests pass on Windows and Linux, including CLI rejection of resealed semantic forgeries, bad decoders/null witnesses, altered samples, incomplete plans, scope promotion and omitted failure/rejection costs.
- All 46 claim-registry and 11 active-surface inventory regression tests pass. Reader style, Lean style/count, axiom consistency, public surfaces and CI-budget checks pass; the native-decide inventory passes on Linux.
- Every registered paper rebuild and its warning gate pass with pinned Tectonic 0.15.0. The changed paper reproduces identical PDF bytes on a second build, its four new-section pages were visually inspected, and the release manifest validates. No release bump.
- All existing 250 claim rows remain unchanged. Only the new mathematical-interface claim extends the reviewed premise-classification projection; no physical premise is discharged.

This closes the finite native-recovery and temporal-feasibility construction gaps. It does **not** derive full M1. Connectivity permits recovery across the whole component; it does not select `L/sqrt(q)`. The existing causal-limit condition also admits other mesoscopic exponents. The particular public meaning, actual A1 response/endogeneity and record interfaces, canonical A3 proposal/cover inputs, physical guard/protection/precision/clock, and identification with M1's semantic read-from order still need source derivations. These results are not a countermodel to the full axioms, and destructive tomography does not replace the version-preserving bank compiler.

### Audit corrections

- Regenerated the active-surface inventory for the new paper fragment; its stale entry caused the original mandatory shard failure.
- Reproduced and fixed a certificate-helper bug that silently ignored extra matrix coordinates. Ragged rows, inexact entries and boolean entries now fail closed; positive zero-dimensional and 729 exhaustive two-column controls remain valid. The packet CLI already derives fixed-size rows independently.
- Made the `HasRoot` prerequisite explicit throughout the contract, summaries, paper and registry: the receiver initially samples its own scalar, then retains samples after admitted means. A two-port erasure control demonstrates why this hypothesis is necessary. Also documented that the Lean response horizon counts sample slots, including the initial sample.
- Exhaustively checked all 38 connected labelled four-port graphs, all receivers and every basis state; disconnected and one-port cases have separate controls. Rebuilt the 98-theorem transitive axiom gate and regenerated all evidence pins and paper artifacts.


muellerberndt (approved):
Reviewed exact head b312639cef5d455afa1857bd9000f759172aad8f, including all 98 public theorem/audit entries, independent executable verification and the paper derivation. The connected recovery construction, joint-kernel continuation criterion, calibrated-receiver guard and charged rejected proposals support the stated finite conditional result. The 15,360-port certificate is correctly bounded to topology; it is not a large payload replay or immutable-version/physical-order identification.

Approved for integration through #963 with the dependent #962. Maintainer changes propagate the result to the flagship and book, remove project-progress narration and renew contract-only byte pins. Exact samples, retained records/arithmetic, physical control, precision, clock, metric read radius and source-law selection remain explicit obligations in #740/#779. Integration is held for exact candidate CI.

PR #964, opened 2026-09-22
Derive native publication and finite A3 source-selection constraints

MarioPoneder:
Finite-deadline projections need not agree under truncation, and neither a selected state nor a successful continuum construction selects M1's source read law. This PR derives a compatible eventual-publication law, permanent native path rank, and the role of the radius regulator. It also proves the support and failure-observable criterion for the full finite noncommutative A3 state objective, with quantitative witness bounds and exact state-versus-process controls.

## Objective

Investigate central source selection, population and radius using the merged recovery, checkpoint-selection and source-net results. The full M1 exit remains a derivation from the complete A1–A3 source architecture; a chosen scheduling interface or successful continuum test does not discharge it.

## Deliverables

- Construct the least harmonic publication potential from finite continuation masses. Derive compatible cylinders, their actual finite-deadline marginal limit, and the path-space information projection. Connected native completion gives a quantitative tail and effective approximation with rational matrix interfaces, an exact rational reference evaluator and a supplied positive rational uniform lower bound.
- Prove permanent ordered-minor positivity on native paths. An analytic Cauchy–Binet extension gives permanent full preparation rank for arbitrary numbers of unknowns at nonadjacent vertices. With retained exact samples and a uniformly positive reference, eventual complete recovery is automatic: no guard or successful-word restriction is needed for this prepared path class.
- Put competing meanings in one declared joint algebra. The selected demand weights are proportional to prior times publication probability; where all meanings publish almost surely, the prior is unchanged. This directly tests the proposed central-selection mechanism without treating the declared algebra as canonical A3.
- Show that every fixed radius exponent between zero and one, and an alternative uniform population, passes the stated continuum tests. Derive a square-root balance from the existing generated-volume error bound, including its bound-dependent coefficient. Removing every spatial restriction collapses the actual generated order and drives causal-interval ordering fractions to one; a fixed axial grid stencil instead produces an L1 cone.
- Prove analytically that the actual finite weighted Umegaki objective preserves every feasible scored local support, without assuming commutation or a global extending state. A positive scored failure effect vanishes at the optimizer exactly when the feasible family excludes it. Derive a failure-probability floor from any feasible witness and an explicit improvement against a false approximate minimum. Treat singular references and positive readout pullbacks separately. Kernel-check the real reductions; do not claim Lean has formalized quantum data processing.
- Independently reconstruct noncommuting rational state segments and four quantum channels. The same stationary state permits perfect record retention or depolarization; the declared full Choi problem selects depolarization, while an added zero-mismatch face selects dephasing. Neither channel family is asserted to be the full A1 instrument grammar.
- Add compact independently recomputed controls, adversarial verification, Lean audit targets and Windows/Linux CI. Integrate the results into the spacetime paper, flagship, book and claim registry, with refreshed previews on release r2045.

## Exit and scientific boundary

The exact radius L/sqrt(q) is unnecessary for the proved continuum target: it is a useful regulator balancing time and spatial approximation errors, not an established physical source length. An appropriate source-derived local propagation law remains necessary. The exact finite M1 menu is a separate obligation.

The finite A3 theorem closes its support-selection question for identified positive scored effects. It does not prove that the actual A1/A2 source grammar excludes M1 failures; conditioning on success or inserting reference zeros would assume that conclusion. The witness floor is not a uniform continuum obstruction.

The PR does **not** derive M1 from complete A1–A3, select its golden population or immutable-version demand, or prove the actual source's propagation/refinement join. The isolated scalar paths and geometric alternatives are not full-axiom countermodels. PR-52 and the source-selection obligations of #740/#779 are not consumed or closed. The contract explicitly retains this unmet full-M1 exit.

## Validation

- 89 new declarations pass the transitive Lean standard-axiom gate. The derivation distinguishes kernel-checked results from analytic measure/entropy, native completion, general exterior-power arguments and the noncommutative matrix entropy proof.
- 103 tests pass on both Windows and Linux, including all 1,962 individually corrupted and resealed scalar fields, missing cases, malformed JSON, stale pins and execution with producer imports blocked. Real CLI tests reject forged receipts and actual proof-file tampering in an isolated copy.
- Independent replay checks 29,523 publication words, 15,128 full-minor path words, 181,217,110 ordered geometric pairs and 18 reconstructed locality graphs, plus three noncommuting state segments and four independently reconstructed quantum channels. The control packet is 30,471 bytes; expanded tapes are recomputed.
- Claim, axiom-inventory, reader-style, Lean-style and theorem-count checks pass. All 252 existing claim rows are unchanged.
- Paper and book builds, warning gates and release-manifest validation pass. Changed PDF pages were rendered and visually inspected.

## Boundary checks

- The rational algorithm explicitly requires an effectively evaluable rational reference and a known rational lower bound. A noncomputable constant reference gives a noncomputable publication potential despite rational native maps and a known positive lower bound; the real-valued existence theorem does not imply an algorithm.
- The deadline-mismatch theorem compares an actual finite conditioned word law at a feasible deadline. An infeasible deadline has no optimizer; Lean's totalized division is not treated as a probability law.
- Recoverable-demand positivity is proved with nonnegative competing potentials, allowing impossible alternatives. A three-demand control checks their zero mass and the positive masses of recoverable labels.
- Source custody includes the imported serialization package initializer. Each scalar-field mutation is checked against independently recomputed expectations, and valid packets succeed before and after the negative tests.

## Maintainer-style audit

Rechecked the finite noncommutative support proof, Pythagorean witness bound, approximate-minimum improvement, singular-reference restriction, and the state-versus-process examples. Reviewed the native/radius arguments, verifier independence and forgery rejection, manuscript and claim scope, source custody, and the corrections made during earlier maintainer integrations. No additional code changes were required in this audit pass. The original full-M1 exit remains explicitly unmet; the finite state-selection theorem does not supply the actual source family or failure-effect extraction.

On final head `eef78cd8ac4d673a9d6d9616947443fa2ffea75a`, all **16 CI workflows completed successfully**: **50 checks passed**, including both mandatory collection gates. Only the schedule-only nightly job was skipped. The branch is conflict-free against main `afed734528edff214c34d4038a64310544df922d`.

CI logs confirm 103 tests on each platform and the same independently verified receipt ([Windows/Linux verification](https://github.com/FloatingPragma/observer-patch-holography/actions/runs/35706875646)), the new standard-axiom reports ([Lean CI](https://github.com/FloatingPragma/observer-patch-holography/actions/runs/35706875609)), and all repository-wide workers and collection gates ([Mandatory Suite](https://github.com/FloatingPragma/observer-patch-holography/actions/runs/35706875725)).


muellerberndt (approved):
Reviewed the finite publication/path arguments, analytic noncommutative support proof and all 89 audited Lean declarations, keeping their scopes separate. The 103 package tests, independent determinant/entropy checks and receipt replay pass. Accepting the conditional results. Integration #966 applies small state-only prose/custody fixes and preserves every prior claim and premise edge. The canonical source family, positive failure-effect extraction and physical propagation/refinement joins remain outside these conclusions; #740/#779 are not closed.

PR #965, opened 2026-09-22
Derive native source reads, recurrent instrument limits and minimal recovery archives

MarioPoneder:
The registered all-port capture has a connected carrier graph but a disjoint scalar port matching. This PR derives its reads, then follows the information and instrument gaps through recurrent quantum extensions and constructive record recovery. It supplies no M1 radius, golden population or successful-history filter.

**M1 exit remains unmet:** these scoped source, instrument and recovery results do not derive A1--A3 selection of M1, and do not prove its non-entailment. They must not close the source-selection obligation.

## Objective
Prove a finite connection from actual twelve-port operations to numeric records, identify the implemented A1--A3 obligations, and determine what additional state can prevent the measured information loss.

## Deliverables
- **Complete native repair/read classification.** Every ideal word is the projection indexed by its visited pair set, with `2^(6N)` maps. Native thresholded pairs commit at most once. The normalization-aware local read algebra has rank 23 after coverage, and positive indistinguishable preparations give binary read error 1/2. Every attempted repair, event commitment, observer-zero snapshot, feedback choice and committed writer edge is independently replayed.
- **Exact instrument obstructions.** The actual no-op band violates affinity; the terminal phase-preserving lift distinguishes equal density ensembles. Neither defines an exact quantum channel under the stated density interpretation. The phase/rotation controls remain, and all 144 native unitary entries have a rational nonzero certificate. Coherent port probes and their time translates generate M12, whose center cannot supply those same port records.
- **Valid, distinguishable quantum extensions.** Permutation twirling and its dephased version are CPTP, agree on immediate scalar means, and disagree on a subsequent native-unitary read by a certified positive amount. Thus the scalar records do not identify a recurrent instrument.
- **Propagation, erasure and coarse-state tests.** The measured extension has an exact graph-distance influence cone for every cycle. Its first cycle erases exactly `5N+1` normalized input dimensions, with explicit strictly positive indistinguishable preparations. Carrier totals fail an exact Markov lumpability test. Influence is not equated with recoverability or a Euclidean radius.
- **Constructive recovery with explicit costs.** A spanning-tree decoder reconstructs every normalized input from deterministic ideal means and `5N+1` non-tree seam differences; rank-nullity proves this linear archive minimal. Independent Gaussian elimination checks the executed rational recovery. A different flagged random-unitary channel recovers every quantum input, retaining `6N` binary flags and the quantum state. Its mean agrees only in expectation; flags alone carry no preparation information. Neither enlarged archive is asserted to be native or a local M1 decoder.
- **Compact reproducibility and publication.** Seven complete captures at N=4,8,16, controls for duration/support level, a 118 KB packet, pinned transitive source bytes, a producer-free standard-library verifier, 32 kernel-checked Lean lemmas, explicit analytic proofs, adversarial tests, one scoped claim/assumption row, and rebuilt paper/flagship/book previews. No expanded routing tapes.

Simulator companion: https://github.com/muellerberndt/oph-physics-sim/pull/16. Linux/Windows CI checks out immutable simulator commit `84d46b97b07965406c608e1465bec18b2a308372` from the fork. The verifier copy is byte-identical to that committed source. Base is main `afed734528edff214c34d4038a64310544df922d`; no unmerged predecessor is required.

## Exit and scientific boundary
The named executable's operation-to-numeric-read map, intervention family, failure witnesses and population/refinement behavior are derived and reproduced. The continuation resolves channel compatibility, recurrent influence versus loss, and exact recovery with quantified additional records. The simulator's native arithmetic and all seven earlier captures are preserved; constructed recurrent channels are explicit controls.

**M1 is not derived.** The positive intervention family is an open preparation interface, not a selected PRNG run or canonical A1/A2 family. Global custody metadata is outside the numeric local channel. The existing finite A3 support theorem needs canonical feasible-family membership and a positive scored failure pullback before it can exclude failed reads. The present source does not supply that architecture, selected public meaning, population, metric radius or compatible refinement. The recovery constructions neither assume nor discharge those obligations, and no full A1--A3 countermodel is claimed.

## Validation
- `Geometry.SourceOperationReadsAxiomAudit`: 32 lemmas; only standard `propext`, `Classical.choice`, `Quot.sound` axioms. The all-time cone, incidence rank/minimal archive and flagged-channel arguments are analytic; their precise boundary from kernel-checked reductions is documented.
- Clean committed simulator checkout: **127 tests pass**, including primitive-source adversarial controls, ordered-port checks and all 39 existing native-capture regressions.
- OPH: **82 tests pass** (36 source-read and 46 claim-registry tests). Empty/forged channels, missing flags/archives, altered outputs, fake circulation, tiny propagator entries and source tampering are rejected.
- Live Linux replay from the immutable simulator commit reproduces the independently verified Windows capture's derived semantics. All seven original cases and original quantum controls remain unchanged.
- Claim registry, axiom inventory, reader style, Lean docstrings/theorem count, diff checks, paper/book builds, warning gates and release manifest pass. Changed PDF pages inspected. Prior 252 claim rows and prior matrix rows are preserved. Previews retain r2045.

Two unrelated Windows frozen causal-history receipt failures are documented in the simulator PR and were reproduced on pristine upstream. No receipt or test was weakened to hide them.

## Audit corrections
- Partial-word projection ranks are checked by building and eliminating their actual matrices: the first three N=4 captures have ranks 40, 32 and 24. Full-sweep rank is reported separately.
- Live reproduction checks every initial, final and recorded ledger value within an explicit absolute 2^-40 comparison budget, as well as source basis and derived semantics. This budget is numerical reproducibility, not a physical precision claim.
- Actual driver writes now consume the instrument primitive's return value. A mutation test proves that changing that value changes committed records and fails the driver's independent replay control. All retained raw captures and controls remain unchanged for the genuine primitive.
- Binary-guess versus abstention error claims and the [0,1] native arithmetic domain are explicit.
- OPH's Linux/Windows source workflow runs the pinned simulator's complete 127-test affected suite in addition to offline and live OPH verification. The simulator upstream reports no separate PR check runs.


muellerberndt (approved):
Reviewed all 15 Lean reductions, the analytic operation/read and phase arguments, all 163 pinned simulator source files, and complete live capture replay. Independent rank, projection, phase and covariance controls pass. Integration #966 applies small even-population/configuration qualifications and a macOS case-alias fix with strict same-file/byte custody; 25 package tests, 29 simulator tests and an independent custody review pass. The named disjoint-pair driver is not promoted to a complete A1–A3 model or a physical spacetime derivation. No full issue exit is discharged.

muellerberndt:
I noticed the new commits through `5ab5e62a7b5c091fdf70ac974d8cc43ef4747051` while integration #966 was in CI. My approval is explicitly attached to the earlier `28aa3840` head; I am reviewing the instrument/recovery delta independently and incorporating it before merging. Please put further feature additions in a follow-up PR while this review finishes, so the reviewed commit and its full-build evidence stay aligned. The finite source-selection obligations remain open as stated in your updated scope.

muellerberndt (approved):
Reviewed the new instrument/recovery delta at this exact head, including all 17 added Lean declarations and their audit entries. Independent rank, archive, branch-recovery and numerical checks passed; 43 integration tests and exact simulator replay passed. The integration retains the updated 2^-40 replay budget and macOS custody guard. Small maintainer corrections clarify even N, connected-graph rank scope and 4 eta < 1/2 in the inverse bound. Approved for integration through #966, subject to its exact-commit CI. This does not identify the constructed extensions with the native/full observer model or discharge the complete #740/#779 contracts.

PR #968, opened 2026-09-22
Derive M1's causal/count limit under one explicit transport law

MarioPoneder:
The current typed A1–A3 data-access clauses do not require a carrier's private central record to reach another carrier. This PR constructs a full-response scalar-seam model and gives the analytic operational, refinement and A3 membership argument, with an all-program remote-read obstruction.

The positive result proves that one explicit proposed source law, **RG: response-natural classical record transport**, suffices for the operational M1 causal/count limit. The complete-response theorem supplies the intrinsic three-dimensional simple ideal; RG makes its rotations act on scalable retained-record transport. Primitive flight and wait composition derives the round access cone. Finite preparation and parallel composition produce the record families, and a sandwich between consumed-writer order and complete native ancestry establishes continuity-set, interval and strict-pair limits, including ordering fraction 1/10. Uniform whole-program timing errors and auxiliary controller dependencies are included. Native ancestry uses localized interventions at intermediate record outputs, holding other registers and cached copies fixed; equality on initial inputs alone does not establish that result.

M1's golden population and radius are compared with a rational-grid regulator of the same operational limit. The radius is the reach of a chosen observation duration, not an asserted uniquely selected microscopic length. The proposed law does not take Fibonacci sampling, a cubical census, a metric read menu or a successful M1 history as input.

RG is a substantial operational extension with explicit workspace, storage, homogeneity, clock, composition, refinement and native-realization requirements. It is not claimed to be minimal, uniquely necessary, derived from A1–A3, adopted as a core axiom, or verified in the registered simulator. The proved endpoint is sufficiency of this complete law for realizable observer experiments and their causal/count limit. Spontaneous microscopic census, compulsory measurement policy, matter dynamics and laboratory calibration are excluded.

The implementation retains compact reproducible evidence rather than expanded routing tapes:

- Independent exact checks of the 12-dimensional response, 66 ordered brackets, 60 same-response closed paths, all 720 covariance identities and 3,600 compositions; full-coordinate grammar and scalar-seam intervention controls. The support checker requires the five-valent seed and reconstructs each oriented midpoint subdivision; equal-count nonicosahedral spheres and altered refinements fail.
[...]

PR #977, opened 2026-09-22
Reduce record gluing to a local clock and derive its information cost

MarioPoneder:
RG currently combines strong continuum transport symmetries with classical programmability and a native realization requirement. This follow-up separates those obligations and proves a simpler sufficient route to the operational M1 causal/count limit. It targets current `main` independently of the unmerged code in #968.

- Finite tetrahedral carrier covariance plus a positive quadratic local-clock expansion forces isotropy. Complete time-faithful refinement excludes coarse shortcuts, while local directional coverage and finite composition attain strict timelike reads. A nonlinear-latency model removes exact scaling and finite null-boundary attainment; a polyhedral control shows that symmetry alone is insufficient.
- A resource-explicit NOT/Toffoli compiler implements finite classical maps with retained inputs and clean work. A finite matrix-algebra construction supports classical copying with twelve central ports and usable quantum seams.
- Actual gate equations define a finite classical A3 history problem whose unique entropy minimizer equals the executed input ensemble. Defective outputs remain feasible: a negative control separates transition admissibility from successful target reads.
- Lossless reads impose an information-cut bound. Dense M1 observation requires at least order q^5 read incidences per fixed horizon and incoming bandwidth at least c*q^2/(2L) per interior receiver under fresh independent interventions. Self records are excluded from incoming transport, multicast is distinguished from unicast, and native event counts require a separate per-event capacity bound. A logical receiver needs a growing composite register bank.

The entropy Hessian supplies an independently motivated candidate quadratic form, but its identification with elapsed time is an additional calibration law. The proposed clock, directional coverage, independent workspace and complete refinement/intervention interface remain explicit physical premises. Finite algebra coherence and the declared history-state problem do not certify an all-level A1--A3 source, native response completeness or the simulator. No core axiom or existing M1 physical gate is marked closed.

The paper, assumption dictionary, two scoped claim rows and dependency/falsification tables agree on these boundaries. `code/rg_principle/AUDIT.md` records the maintainer-style review. The receipt stores compact hashes/counts; verification recomputes traces rather than committing raw tapes.

The follow-up audit independently binds every named reference circuit to its intended function and the retained-copy diagnostic to its intermediate writer. Coordinated circuit/table substitutions, cached-writer substitutions and corrupted submitted Kraus operators fail verification. The tracial minimum requires local references to be restrictions of the same joint reference; an incompatible-reference counterexample demonstrates why. Two inherited CI failures are corrected: a stale source-operation receipt pin after a comment-only Lean edit, and Windows newline translation inside a strict CSV fixture. The receipt is regenerated with its scientific results unchanged, and the production CSV validator remains strict.

Validation: 67 RG tests pass; the affected suites total 232 passing tests on Windows and 167 on Linux. The receipt replays 263 circuits, 442,496 forward/inverse/intervention/comparison gates, 7,880 localized interventions and 40 paired writer probes, plus exact clock, Kraus, full-history and resource controls. Twelve Lean theorems pass a transitive standard-axiom whitelist audit with negative proof-shortcut controls; the source-operation audit target also builds. Repository claim, axiom, prose, Lean-style, CI-budget and receipt-portability gates pass. The paper/book build, warning gate and release manifest pass; the revised appendix was visually checked.


muellerberndt:
Heads-up on two overlaps now that the small-PR batch has landed on `main` (ba84976a).

**The Windows CSV fixture fix is now on main.** `main` was red before this batch on `mandatory shard (windows-latest, 7)`, and I fixed it independently at 68a05898 with the same `newline=""` change you made in a722fdf0. Your version is the better one: the extra `cell_break` parametrization over `["\n", "\r\n"]` also pins that a CRLF *inside* the cell round-trips, which mine does not cover. When you rebase, please keep your parametrization and drop the duplicated write-call edit.

**Rebase surface.** This PR was `CLEAN` against the old `main`; against the new tip it conflicts on three paths:

- `tools/test_gates_actually_fail.py` — the duplicate fix above.
- `paper/paper_release_manifest.json` — regenerate with `python3 tools/generate_paper_release_manifest.py --preview` rather than resolving by hand.
- `paper/recovering_observer_spacetime_and_einstein_dynamics_from_overlap_consistency.pdf` — binary; take the rebuilt artifact. `paper/tex_fragments/PAPER.tex` changed in the batch (a capacity-coordinate digit sync), and that paper inputs the fragment, so the PDF genuinely moved.

Two gates worth running locally before pushing, since neither is reachable from a fork PR check:

- `python3 tools/check_reader_style.py` — local-only; it rejects em dashes and internal register identifiers in reader prose. Three PRs in this batch tripped it.
- `python3 -m pytest code/audit/test_public_claim_semantics.py` — the README case section has a per-bullet word budget, and its split folds the closing paragraphs into the last bullet. English headroom is currently 2 words.

Nothing here is a review of the research content of this PR.

muellerberndt:
Merged into `main` at 7f588ea5 through the integration branch `integration/prs-20260923` (merge commit c2fe8526), on top of PR 969 (3a588528) and PR 968 (12e98f1c).

**Review outcome.** The twelve Lean statements build locally with only `propext`, `Classical.choice` and `Quot.sound`; the receipt replays exactly (263 circuits, 442,496 gates, 7,880 localized interventions, 40 paired lowering probes); the two claim rows, the assumption dictionary and the appendix section state the same boundary as the derivation: LC1–LC3 are proposed physical premises, the isotropy/cone/capacity results are conditional mathematics, and no all-level A1–A3 source, native response completeness or simulator property is certified. The interior-cube and bandwidth arithmetic (`3m² < n²`, `n ≤ 2m+1`, `q-2m ≥ n²/2`, bandwidth `c q²/(2L)` after excluding the receiver's own record) checks out. Gates 740 and 779 stay open; no premise is discharged.

**Maintainer changes in 7f588ea5.**
- `code/rg_principle/AUDIT.md` was removed: the workspace placement policy keeps review logs and audit reports out of the scientific repository. Its content is reproduced below so nothing is lost.
- The README, CONTRACT and DERIVATION no longer describe PR 968 as pending or unmerged; they point at the RG statement in `code/source_selection_model/RECORD_GLUING.md`. The receipt was rebuilt for the repinned documents (scientific fields unchanged).
- The appendix section now cross-references the RG definition (`Section~\ref{sec:source-record-gluing}`), which lives in the same paper after PR 968.
- The premise-dependency projection hash was recomputed for the union of the four new claim rows from PRs 968 and 977, and `claims/active_surface_inventory.json` was regenerated.
- Your Windows CSV fixture parametrization was kept over the duplicate `newline=""` fix on `main`.
- The spacetime paper and the book were rebuilt, the preview manifest regenerated, and the README theorem-count floor lifted to 12000 (the merged library has 12,008 public declarations; neither PR crossed the floor alone).

<details>
<summary>Relocated review text (formerly <code>code/rg_principle/AUDIT.md</code>)</summary>

## Maintainer-style scope and correctness review

This review covers the local-clock reduction and its finite controls against
`main` commit `992486deb21dd64fd58a4afb2056a431327e2e59`, including the
maintainer's registry-scope and CI-cache corrections. The work is independent
of the unmerged implementation in PR #968.

### Claims examined

* **Cone selection:** finite symmetry fixes an invariant quadratic form;
  it does not create local directional coverage. Uniform quadratic regularity,
  operational attachment of the response ideal and complete time-faithful
  refinement are explicit premises. The proof includes all control/timing
  dependencies. The nonlinear-clock example removes exact scaling without
  claiming finite null-boundary attainment.
* **Coherence:** the twelve-central-port tensor model and copy channel are
  finite witnesses. An explicit joint observer resolves state compatibility
  for that finite family. Neither a dormant support complex nor a prescribed
  response subalgebra proves complete A1--A3 membership. These conclusions
  are excluded in the paper, contract and claim registry.
* **A3 join:** actual gate equations define a classical history state problem.
  Conditional relative entropy reproduces the input ensemble. A deliberately
  defective program remains feasible and fails the separate truth-table
  check. No successful-read condition defines the feasible family.
* **Programmability:** NOT/Toffoli suffice. The constant-control bit used to
  lower CNOT is prepared, cleaned and charged. Inputs and all work registers
  are checked, including nonblank output registers. This is a constructive
  all-size analytic argument with a bounded executable reference suite.
* **Writer identity:** the cached-version control agrees on every initial
  input and fails isolated intermediate interventions. Initial equivalence
  is not treated as evidence of correct writer ancestry.
* **Capacity:** the bound uses a complete finite transcript and independent
  interventions while caches are fixed. Timing side channels are counted.
  Incoming bandwidth excludes the receiver's own locally available record;
  the audited bound is c*q^2/(2L), not c*q^2/L. Aggregate read incidences are
  not unicast transmissions. Native event lower bounds require an additional
  per-event capacity bound. A logical receiver needs a growing composite
  register bank; one C^12 central record has only twelve alternatives.

### Verification coverage

The reference receipt recomputes all 263 circuits: 442,496 forward, inverse,
intervention and boundary-comparison gate executions, 7,880 isolated
intermediate interventions and 40 paired lowering probes,
all 1,728 basis units of the copy channel, exact clock/entropy calculations,
the complete 64-history Toffoli state problem and integer resource controls.
It stores hashes and counts, not the raw tapes. The checker does not import
the compiler. Its feasibility predicate does not test the desired truth table.

Hostile tests reject malformed types, out-of-range/duplicate operands,
incomplete tables, dropped/retargeted gates, dirty work, cached substitution,
stale source pins, missing cases, forged receipt quantities and a forged
entropy prior. JSON duplicate keys and nonfinite constants are rejected;
optimized Python cannot bypass the certificate checks.

The twelve Lean statements have a transitive axiom whitelist audit. Negative
controls reject both `sorryAx` and compiler-trust proof shortcuts. The Lean
coverage is algebraic/path/limit/capacity mathematics; the differentiability,
general compiler induction, matrix-state and entropy arguments are analytic
proofs in the derivation, not claimed as fully formalized Lean results.

### Exit assessment

The contract's finite coherence, simplification, countercontrols, consequences
and reproducibility deliverables are supplied. The remaining physical work
is the native source completion and testing of the proposed clock/refinement
law, not a withheld step of the stated conditional proofs. No empirical RG
validation, canonical axiom amendment or complete source-selection theorem
is asserted by this PR.

### Follow-up audit corrections

The reference receipt previously checked each producer-supplied truth table
but did not independently enforce the meaning and completeness of its named
catalog. The checker now specifies all 263 target functions independently,
rejects duplicates/missing names and rejects coordinated circuit/table
substitution during receipt recomputation. The retained-copy entry also
checks its named intermediate writer against a separate boundary contract;
initial truth-table agreement cannot bypass this check. The additional
comparison gates and probes are counted.

The copy-channel validator accepts actual submitted sparse Kraus operators
and rejects dropped, duplicated, source-erasing and wrong-buffer operators.
An independent producer remains responsible for constructing the candidate.

The tracial-minimum proof requires local references to be restrictions of
the same joint reference. Faithfulness and traciality without compatibility
do not suffice: an exact two-reference counterexample has nonzero entropy
derivative at the proposed joint minimum. The paper and assumption dictionary
state the compatible-reference premise. The latency argument states its
nonnegative grading, zero-displacement waits and inheritance of quadratic
symmetry explicitly. Inner-menu resource bounds are distinguished from the
all-n bound for the full-radius experiment.

CI exposed two inherited integration failures. The source-operation receipt
still pinned the old `ChoiCPTP.lean` comment bytes after main's documentation
correction. A full recomputation changes only that source hash; the native
derivation and every other receipt field remain identical. The Windows CSV
fixture translated embedded LF text to CRLF, so the strict registry-scope
comparison correctly rejected its changed content. The fixture now preserves
bytes and tests both LF and CRLF cells; the production validator is unchanged.

Local follow-up validation passes 232 affected tests on Windows and 167 on
Linux, including 67 RG tests. Both the record-gluing and source-operation
Lean audit targets build. Registry, axiom consistency, reader/Lean style,
CI-budget and receipt-portability gates pass. The changed paper rebuilds
with no overfull boxes or reference/citation/glyph/font problems; its preview
manifest validates and appendix pages 183--186 were visually checked.

</details>

PR #981, opened 2026-09-23
Prove sparse M1 replacements with preserved entropy and full free-field limits

MarioPoneder:
M1's complete metric-neighbor menu is stronger than several of its downstream continuum conclusions require. This PR gives explicit sparse replacements, proves which named observables they preserve, and supplies counterexamples and lower bounds showing where a weaker read law fails. It includes main `7f588ea5`, including the merged source-selection and RG work.

The objective and exit are recorded in `code/m1_necessity/CONTRACT.md`. Every comparison graph, tick, cell weight, action and route is specified before testing; successful propagation or entropy behavior is not an assumed selection criterion.

### Completed results

- Construct charged integer paths for sparse S and A stencils and prove their round causal limit, actual generated-diamond and strict-pair limits, ordering fraction 1/10, retrospective count clocks and supplied conformal mass identities. Explicit alternative population/radius schedules establish that these conclusions do not uniquely require Fibonacci populations or the square-root radius.
- Prove that this agreement alone does not preserve crossing entropy: S's crossing ratio to the dense graph vanishes, while A retains a diagonal/coordinate density ratio of sqrt(2) that scalar recalibration cannot remove.
- Construct T, which also preserves the original leading crossing coefficient in every fixed regular interior planar orientation, with exactly one nat per unordered crossing pair. At fixed q it has degree Theta(q^(9/8)) and fixed-horizon logical reads Theta(q^(17/4)), versus Theta(q^(3/2)) and Theta(q^5) for the dense comparison. Its larger vanishing radius converges more slowly; this is not an equal-accuracy runtime claim. A complementary theorem proves that original-radius symmetric translation-invariant thinning preserving the three coordinate-cut leading counts must keep asymptotically almost every vector.
- Find and fix a substantive defect in T's first action: its uniform weights collapse m^3 modes toward the mass frequency, introducing m^3-1 extra low-energy modes and divergent thermal energy. The final action is the fixed positive half-sum with nearest-neighbor energy on existing reads. It retains every coupling and the crossing statistic while supplying uniform control of the whole spectrum.
- Prove ordered Dirichlet eigenvalue convergence with multiplicity, a uniform j^(2/3)/(2L^2) spectral lower bound, full thermal partition/normal-ordered energy/Gibbs entropy limits, and smooth compact free quantum Weyl and causal-commutator limits. The limiting quantum response and read/count cones agree, including observations rounded to declared ticks. The clipped box permits zero mass through its proved boundary gap. Separately, prove the centered-leapfrog obstruction at the radius/c tick for the positive isotropic action class, including the dense graph.

### Necessity audit and evidence

The inventory binds all eight direct consumers and 95 existing claims in their registered dependency closure, preserving each row's assumptions and dependency roles. All 262 existing main registry rows remain unchanged. Four new mathematical comparison claims are registered without discharging physical source-selection premises. Exact finite tuples, native execution, historical instruments, interacting matter and gravitational identification retain their original contracts. Thermal Gibbs entropy and the crossing statistic are distinct observables.

The receipt is about 215 KB; there is no raw history archive. Independent verification reconstructs stencils, routes, cuts and actions without importing the producer, and replays 149,760 event instances and 13,479,912 read incidences. Hostile controls reject empty/truncated catalogs, forged outputs, omitted interventions, legal but out-of-tube detours, missing stabilizing weights and hidden alias modes. Tests cover every t=1 periodic alias mode and the entire spectrum of a small clipped operator. Large T populations are not claimed fully executed.

Seventeen finite reductions are kernel checked with transitive standard-axiom audits rejecting `sorryAx` and compiler-trusted proofs. The universal geometric, measure, spectral and quantum arguments are analytic proofs; the documentation does not describe them as fully Lean formalized.

### Audit follow-up

The audit report is kept outside the scientific checkout, following the maintainer's placement correction on #977. The spectral proof now explicitly requires bounded continuous functional calculus, with projection cutoffs outside the limiting spectrum; the stated covariance and evolution functions satisfy those conditions. Resealed T-route controls independently reject wrong endpoints, excessive charged time, out-of-tube legal detours and false commitments. An additional 408 half-tie, corner and deterministic randomized routes cover every supported T level, including levels 7 and 8 beyond the committed reference catalog.

### Validation

- All 55 executed CI checks passed on `eabfc396`, including every Linux and Windows mandatory shard, Lean, certificate suites and publication checks. The nightly-only job is skipped by workflow design.
- The subsequent main change `9b527a4f` adds README text only. Its combination with this PR passes reader style, active-axiom inventory and all 31 public-claim tests; the PR remains conflict-free.
- After the audit corrections, all 80 replacement-package tests passed on Windows and Linux. Reader style, public-claim semantics, registry and publication-manifest validation passed again.

- The 153-step standard suite passed before the latest-main integration (151 steps on Windows; the last two unchanged gates on Linux because Windows denied symlink privilege). After integration, all five newly added mandatory steps passed, including 35 tomography mutation tests and two subtests. Additional heavy and optional exact-certificate lanes are outside this standard run.
- After integration, 387 affected tests passed on Windows and 387 on Linux, including 75 replacement-package tests and the inherited source-operation receipt checks.
- `Geometry.M1NecessityAxiomAudit` built successfully for all 17 public reductions.
- Registry, reader/Lean documentation style, active axiom surfaces, public-claim semantics, Lean CI budget and publication-manifest checks passed.
- The paper was rebuilt with no overfull boxes or reference, citation, glyph or font problems; the new appendix and reference transition were rendered and visually reviewed.
- The merged source-selection and record-gluing receipts were independently replayed; only the Lean CI workflow pin changed. The source-current inventory was regenerated to include the two new Lean files and those receipt hashes; every scientific classification and conclusion is unchanged. Its independent verifier and pre-generation source-lock gate passed.

The exit is a completed replacement theorem for the named observables, together with concrete failures and an original-radius obstruction. It does not claim that an explicitly defined regulator is already selected by OPH's physical axioms.


muellerberndt:
Merged into main together with the three sibling M1 packages, through an integration branch: this PR's merge commit is c353f416, followed by the merges of the interface, vacuum-fidelity and transport PRs and one maintainer commit 1f37f2e8 that carries the fixes and the regenerated custody for the union.

Audit outcome: no blocking findings. The stencil, cut, route and spectral statements were spot-checked independently (axial cut formulas against brute-force pair enumeration, the sqrt(2) density ratio, the 9/8 and 17/4 exponents, the alias floor and the j^(2/3)/(2L^2) bound), the 80 package tests and the verifier pass, and the receipt regenerates byte-identically at the PR head.

- `CONTRACT.md` and `CONSUMERS.md` no longer narrate the integrated base commit or pull-request numbers; `CONSUMERS.md` says "later" instead of "in a later PR". Package documents in this repository describe the science and its scope, not the project state.
- Appendix wording: "The profile and its physical interpretation are supplied" now reads "are inputs, not results of the comparison"; "consumer audit" is "consumer inventory"; the closing sentence now says "this comparison establishes no physical source-selection premise" instead of "no ... premise is declared closed".
- `test_necessity.py` reads `lean-ci.yml` with `encoding="utf-8"` (the workflow contains U+00B7; cp1252 happens to decode it, so this is latent).
- `Geometry.M1NecessityAxiomAudit` is imported by the `Lean/Geometry.lean` umbrella like every other audit module.
- Integration consequence: the downstream closure of the eight M1 consumer rows now reaches the thirteen comparison claims added by the three sibling PRs, so `downstream.json` binds 108 rows instead of 95, and the count in `README.md`, `CONSUMERS.md` and the test pin follows. The receipt was rebuilt after these edits.

One editorial follow-up for a later pass, not a request now: the spaced-ball stencil is now defined in two appendices (this one and the interface appendix), and the alias-mode defect has two completions (the half-sum action here, the short-ball repair there). Both are correct as stated; consolidating them into one definition and one comparison would shorten the paper.

PR #982, opened 2026-09-23
Prove variational area and uniform-action thermal limits for sparse M1 reads

MarioPoneder:
Fixed smooth-section area agreement can conceal microscopic partitions with the wrong entropy and a uniform field action with extra low-frequency modes. This PR constructs and proves an explicit sparse repair that preserves the full variational perimeter observable, its finite-volume minimizers and the complete free thermal limit.

The central result is a shared sharp threshold. Starting with the spaced-ball stencil `T`, add the complete short integer ball `B_r` with unchanged one-nat crossing weights. Writing `s=r/K`, bounded-energy partitions are compact precisely when `s` tends to infinity within this repair class. The same condition removes the extra oscillator sector of the uniformly normalized action. Compactness and the all-mode estimates are proved from the actual reads.

The results are:

- **An explicit obstruction.** Half-residue partitions have exact half volume and crossing entropy tending to zero on the torus. Localized inside a clipped cube, they fall below the Euclidean isoperimetric value by the factor `2^(-1/3)`. Exact finite cardinality correction costs a vanishing amount. Thin connecting paths preserve the noncompact counterexample, so connectivity alone does not repair it.
- **A complete constructive repair.** The schedule `q=2^(16t), m=2^(8t), K=2^(6t), r=2^(7t)` proves bounded-energy compactness, the full strong-L1 perimeter Gamma limit, binary recovery and exact finite volume. Finite minima converge, and every sequence of normalized near-minimizers has subsequences converging to continuum perimeter minimizers. Where the prescribed ball fits inside the allowed interior cube, the limiting minimum is `(36*pi)^(1/3)*v^(2/3)` and every limiting minimizer is a ball.
- **The full spectral classification.** At `s -> gamma` in `(0,infinity)`, the complete positive-mass thermal limit has eigenvalues `4*pi^2*(|j|^2 + gamma^5*|ell|^2)`, with two integer triples. This is the oscillator spectrum of a product of two three-tori, describing regulator spectral multiplicities. At `s -> 0`, normal-ordered thermal energy diverges. At `s -> infinity`, the extra sector disappears. The explicit repair also proves clipped Dirichlet eigenvalue, thermal and smooth Gaussian quantum limits using one uniform positive action coefficient, without an added specially weighted nearest-neighbor term.
- **Retained kinematics and explicit cost.** The original routes preserve the local causal/count limits on interior, non-wrapping diamonds. For `P` oriented pairs and `N` interval events, `P/N^2 -> 1/20` and `2P/[N(N-1)] -> 1/10`; truncated or wrapping intervals need their own volume formulas. Read incidences over a fixed horizon grow as `q^(71/16)` rather than the dense construction's `q^5`, at fixed population parameter `q`. Its larger radius converges more slowly, so this is not an equal-accuracy or native resource comparison.

The universal BV, isoperimetric and spectral proofs are analytic in `code/m1_interfaces/DERIVATION.md` and the new paper appendix. Lean checks eight finite cut-change and scale reductions, with transitive standard-axiom audits. Three new claim records preserve all 262 existing records. The defined cut minimum does not identify a physical horizon, and crossing entropy remains distinct from Gibbs entropy. No native source or physical M1/RG premise is marked discharged.

Validation:

- All 55 executed CI checks pass on `df839e48`, including every Linux and Windows mandatory shard and both aggregate acceptance gates. The nightly-only job is skipped as designed; no checks are pending or failed.
- 44 package tests pass on both Windows and Linux; the combined Windows package and claim-registry control run passes 95 tests.
- Independent reconstruction covers all 96 cut cases on 12 graphs and 3,340,024 read incidences. Exact column certificates cover the three larger moment regimes without storing or claiming execution of their full populations.
- Hostile tests reject incomplete catalogs, changed labels, forged counts, false connectivity and spectral certificates, duplicate parent claims, duplicate keys and nonfinite values. Actual shifted-array field actions are checked against independently enumerated edge energies and fixed-zero exterior boundary potentials. Coordinated substitution of a clipped graph Laplacian is rejected even after recomputing its reported totals. Real `python -O` invocations reject spectral and boundary-action forgeries; the verifier also succeeds with producer imports blocked. A CI guard requires replay triggers for every pinned source dependency.
- All eight Lean reductions and their axiom audit compile without warnings. Claim registry, active-surface consistency, reader style, Lean CI budget and compiler-trust inventory checks pass.
- The paper PDF was rebuilt and visually checked, with no unexplained layout/reference/font warnings and a valid preview manifest. Inherited source-model receipts were fully replayed; only their Lean CI custody changed. The Stage-2 inventory was regenerated and independently verified with every scientific classification unchanged.

Built directly on main `9b527a4f`; the discussion extends #981 but imports no unmerged code or theorem from it. The commits separate the proofs, independent evidence, publication integration and audit corrections.


muellerberndt:
Merged into main together with the three sibling M1 packages, through an integration branch: this PR's merge commit is fa18e793, with one maintainer commit 1f37f2e8 carrying the fixes and the regenerated custody for the union.

Audit outcome: no blocking findings. The 2^(-1/3) isoperimetric deficit, the compactness threshold, the 4 pi^2 (|j|^2 + gamma^5 |l|^2) alias spectrum, the q^(71/16) read exponent and the 1/20 and 1/10 pair fractions were reproduced independently; the 44 package tests and the verifier pass, and the receipt regenerates byte-identically at the PR head.

- `README.md` and `CONTRACT.md` no longer mention the base commit, pull-request numbers, "unmerged" code or review placement; where the text referred to the sibling comparison it now names `code/m1_necessity` by path. `DERIVATION.md` says "as proved in" rather than "as proved on main in".
- `Geometry.M1InterfacesAxiomAudit` is imported by the `Lean/Geometry.lean` umbrella like every other audit module.
- The receipt was rebuilt after these edits; the scientific fields are unchanged.

The appendix itself needed no change.

PR #983, opened 2026-09-23
Prove global quantum fidelity and native clock limits for sparse reads

MarioPoneder:
Stable scalar evolution and agreement on smooth probes can conceal a large total quantum error across the unresolved modes. This PR proves sharp global thresholds against the original vacuum and Hamiltonian, supplies an explicit sparse program whose total produced energy vanishes, and proves exactly why its numerical tick cannot also be the proposed native causal clock.

The results are:

- **Full-population fidelity thresholds.** Exact mode formulas and matching time-average lower bounds show that order `p=2,4` needs `N*tau^(2p)*sigma^(2p)` to vanish for global vacuum fidelity, and `N*tau^(2p)*sigma^(2p+1)` for total excess-energy fidelity. Every mode and every time in the interval is retained. Global vacuum fidelity controls every bounded effect; bounded-energy coherent inputs in a fixed band have an additional explicit phase/displacement bound with coefficient one.
- **A constructive quantum repair with accounted cost.** On the fully specified sparse stencil, the actual bulk frequency is `sigma=Theta(q^(7/32))`, not inverse read radius `q^(1/8)`: the short ball dominates degree while the spaced ball dominates the second moment. The standard fourth-order symmetric composition at `tau=q^(-5/8)/64` has total produced energy `O(q^(-1/32))`, global infidelity `O(q^(-1/4))`, and `Theta(q^(79/16))` ordered reference read incidences. The second-order program at the same tick instead has mean energy `Theta(q^(51/32))`. Matching energy-tolerance costs are derived for both sparse and dense families.
- **An attained signal front and a precise clock obstruction.** Before wrapping, the fourth-order canonical support radius `(3j+1)*a` is attained by an exact nonzero influence. A local displacement and bounded sine effect can distinguish it. On the explicit repair, the attained distance can tend to `3/8` while the numerical elapsed time tends to zero. A faithful implementation with a complete fixed-speed influence law therefore cannot identify this simulation tick with the LC/RG native clock. This is compatible with the proved restricted-preparation fidelity; it is not an assertion that all operation grammars fail.
- **An exact-preservation obstruction.** A fixed-degree scalar polynomial symplectic update preserving the original vacuum satisfies a polynomial identity whose degree parity forces the trivial near-identity map once the spectral cardinality is large enough. The required unbounded cardinality is proved for the explicit families. The approximate repair does not silently replace the original vacuum with a modified one.

The integrator itself is standard and attributed to Yoshida. The new contribution is the full-population error theorem, its sparse implementation and precision-dependent costs, and the attained-signal comparison with the proposed native cone law. The classical writer ledger is a reference calculation; the quantum map factors into commuting two-mode phases and onsite unitaries without copying unknown quantum fields. This does not construct a native A1--A3 source, finite twelve-port oscillator implementation, interacting matter law or physical clock. Existing physical premises are unchanged.

Validation:

- 57 package tests pass on Windows and Linux; the combined Windows package and claim-registry/semantics run passes 108 tests.
- All 55 applicable CI checks pass on audited commit `fd52952a`, including both mandatory collection jobs and every shard. The scheduled nightly full-suite job is skipped by design on pull requests. The branch has no merge conflicts.
- Independent replay checks four complete spectra over 523 site instances, 512 population observations representing 66,944 mode-time evaluations, 24 full spatial covariance comparisons, 702 forward and 702 inverse reference writes, and six complete signal rays. Exact cubic arithmetic verifies order cancellation, a common stability domain, inverse execution and nonzero extremal coefficients.
- Hostile controls reject coordinated false-vacuum substitution, omitted or altered modes and times, changed inverse gates with recomputed outputs, false zero signal fronts, forged cost powers, duplicate parent claims, stale pins, malformed JSON and nonfinite values. Actual `python -O` processes reject the substantive forgeries; verification also succeeds with producer imports blocked.
- Stability enclosures are explicitly bound to the exactly multiplied gate coefficients. Identity and doubled-step substitutions are rejected even though they remain symplectic. Forged stale writer histories are rejected even with unchanged final outputs and recomputed trace commitments.
- Ten Lean reductions and their transitive standard-axiom audit compile without warnings. Registry and active-surface checks preserve all 262 prior claim rows and add four scoped results. The paper is rebuilt and visually reviewed with its preview manifest.
- The first `q=65536` certificate uses exact integer-ball columns. It does not execute that full population or stand in for the universal analytic proofs. Evidence remains compact, with no raw large-population archive.

Built directly on main `9b527a4f`, independently of the still-open #981 and #982. The stencil is defined again here; no unmerged theorem or file is imported. Commits separate the proofs, verification and publication integration.


muellerberndt:
Merged into main together with the three sibling M1 packages, through an integration branch: this PR's merge commit is ce7ff531, with one maintainer commit 1f37f2e8 carrying the fixes and the regenerated custody for the union.

Audit outcome: no blocking findings. The fourth-order composition entries, the stability domain, the single-mode excitation and fidelity formulas, the threshold exponents, the q^(7/32) bulk frequency, the q^(-1/32) and q^(-1/4) repair scalings, the (3j+1)a support radius and the parity obstruction were reproduced independently; the 57 package tests and the verifier pass, and the receipt regenerates byte-identically at the PR head.

- `CONTRACT.md` and `README.md` no longer mention the base commit, a pull-request number, "unmerged" files or audit-log placement; the sibling stencil is referred to as `code/m1_interfaces` by path.
- Appendix: after "it does not assert that the quantum approximation fails" the text now adds "Other operation grammars and implementations are outside this obstruction", which the derivation and the claim row already state; the paper now carries the same scope for the clock proposition as for the polynomial obstruction.
- `test_fidelity.py` reads the two workflow files with `encoding="utf-8"`.
- `Geometry.M1VacuumFidelityAxiomAudit` is imported by the `Lean/Geometry.lean` umbrella like every other audit module.
- The receipt was rebuilt after these edits (it pins the documents, the test file, the appendix and `lean-ci.yml`, whose target list now names four audits).

